PR #4, opened 2020-09-24
Add TablePrinter

Tabrizian:


deadeyegoodwin (changes requested):

CMakeLists.txt
```diff
@@ -42,6 +42,37 @@ add_library(
   TritonCommon::triton-common-json ALIAS triton-common-json
 )
 
+add_library(
```
deadeyegoodwin, line 45: You put this in the middle of the triton-common-json setting. Move it all to just before install

src/table_printer.cc
```diff
@@ -0,0 +1,186 @@
...
+#include <numeric>
+#include <sstream>
+#include <string>
+#include <sys/ioctl.h>
+#include <unistd.h>
+#include <vector>
+
+#include "triton/common/table_printer.h"
+
+namespace triton {
```
deadeyegoodwin, line 39: These should be on one line. Did you run the formatter?

src/table_printer.cc
```diff
@@ -0,0 +1,186 @@
...
+#include <iomanip>
+#include <memory>
+#include <numeric>
+#include <sstream>
+#include <string>
+#include <sys/ioctl.h>
+#include <unistd.h>
+#include <vector>
+
+#include "triton/common/table_printer.h"
```
deadeyegoodwin, line 37: This should be first include above, separated by blank line. look at other *.cc file for coding style.

include/triton/common/table_printer.h
```diff
@@ -0,0 +1,78 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#pragma once
+
+#include <memory>
+#include <sstream>
+#include <string>
+#include <sys/ioctl.h>
+#include <vector>
+
+namespace triton {
```
deadeyegoodwin, line 34: one line... run formatter

include/triton/common/table_printer.h
```diff
@@ -0,0 +1,78 @@
...
+  std::vector<std::vector<std::string>> data_;
+  // terminal size
+  struct winsize terminal_size_;
+  // Fair share of every column
+  std::vector<float> shares_;
+
+  // Update the `shares_` such that all the excess
+  // amount of space not used a column is fairly allocated
+  // to the other columns
+  void calculate_fair_share_();
```
deadeyegoodwin, line 65: Why are these functions named like this. Please follow coding convention.

include/triton/common/table_printer.h
```diff
@@ -0,0 +1,78 @@
...
+public:
+  // Insert a row at the end of the table
+  void InsertRow(const std::vector<std::string> &row);
+
+  std::unique_ptr<std::string> PrintTable();
+
+  static void Create(std::unique_ptr<TablePrinter> *table_printer,
+                     const std::vector<std::string> &headers);
+
+private:
```
deadeyegoodwin, line 50: Private functions go before private members.

include/triton/common/table_printer.h
```diff
@@ -0,0 +1,78 @@
...
+// An ASCII table printer.
+//
+class TablePrinter {
+public:
+  // Insert a row at the end of the table
+  void InsertRow(const std::vector<std::string> &row);
+
+  std::unique_ptr<std::string> PrintTable();
+
+  static void Create(std::unique_ptr<TablePrinter> *table_printer,
```
deadeyegoodwin, line 47: Why have a create function instead of just a constructor. We use Create in core triton because it allows us to return an error code but that is not needed here.

include/triton/common/table_printer.h
```diff
@@ -0,0 +1,78 @@
...
+
+//
+// An ASCII table printer.
+//
+class TablePrinter {
+public:
+  // Insert a row at the end of the table
+  void InsertRow(const std::vector<std::string> &row);
+
+  std::unique_ptr<std::string> PrintTable();
```
deadeyegoodwin, line 45: Using a unique_ptr here seems unnecessary. Just std::move() the string on return from the function.

deadeyegoodwin (changes requested):

src/table_printer.cc
Tabrizian: Interestingly it seems like my formatter is breaking them apart. I am using the exact config mentioned in the Doc.

deadeyegoodwin (approved):

PR #17, opened 2021-05-19
Add host policy field

GuanLuo:
Note that the numa id for GPU instance will always be the same as the gpu id, this is because in proto3 there is no way to tell if a primitive type is set explicitly or the default value is used, i.e. 0 can either be the user sets the id to 0 or the user doesn't set the value. Which will be confusing for GPU instance because the default behavior for GPU instance is that instances on different GPU device will be assigned to different numa id, but we can't differentiate the meaning of 0 here.

If we really want to be able to overwrite numa id for GPU instances, a work around is to wrap the numa id in a protobuf message so we will check if the message is specified to determine if the numa id is set explicitly.

### DIFF
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -287,7 +287,7 @@ message ModelInstanceGroup
   //@@     for instance, KIND_CPU is "cpu", KIND_MODEL is "model" and
   //@@     KIND_GPU is "gpu_<gpu_id>".
   //@@
-  string host_policy = 8;
+  string host_policy = 9;
 }
 
 //@@
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -279,6 +279,14 @@ message ModelInstanceGroup
   //@@     not be added to the scheduler. Default value is false.
   //@@
   bool passive = 7;
+
+  //@@  .. cpp:var:: int32 numa_id
+  //@@
+  //@@     The id used to associate with NUMA related features for the instance
+  //@@     group. For KIND_GPU, the default value will be the same as the GPU id
+  //@@     of the instance. For other kinds, the default value is 0.
+  //@@
+  int32 numa_id = 8;
 }
 
 //@@
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -283,8 +283,9 @@ message ModelInstanceGroup
   //@@  .. cpp:var:: int32 numa_id
   //@@
   //@@     The id used to associate with NUMA related features for the instance
-  //@@     group. For KIND_GPU, the default value will be the same as the GPU id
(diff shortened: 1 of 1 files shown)


protobuf/model_config.proto
```diff
@@ -239,6 +239,15 @@ message ModelInstanceGroup
   //@@     not be added to the scheduler. Default value is false.
   //@@
   bool passive = 7;
+
+  //@@  .. cpp:var:: int32 numa_id
+  //@@
+  //@@     The id used to associate with NUMA related features for the instance
+  //@@     group. For KIND_GPU, this field will be ignored and the value will be
```
deadeyegoodwin, line 246: Instead of ignoring, make it an error to explicitly specify numa_id for KIND_GPU.

deadeyegoodwin (changes requested):

protobuf/model_config.proto
GuanLuo: After some thinking, we should be able to have a "device" field, which refers to NUMA specific ID for CPU instance and GPU device id for GPU instance, instead of "numa_id". This was discussed before, but actually the backward compatibility concern may not apply here:

There are actually two notions: "memory type and id" that describes the memory property and "instance kind and id" describes the instance property. I mixed them together before because they are the same in GPU case. And the backend compatibility concern applies to the memory description where one may implements a different frontend / backend that relies on the "memory type and id" even in CPU case, so we shouldn't add more meanings to it.

But for the "instance kind and id", allowing CPU instance to have id different from 0 shouldn't break any existing code. Because the return value of `TRITONBACKEND_ModelInstanceDeviceId()` is not in control of the user and it is always 0 for CPU instance, it is not likely that there is logic that depends on a constant value.

That being said, we still need to extend the memory related APIs so it may return requested memory from location closer to specified device. The difference is that the device will be specified by "device kind / device id" pair but not "device kind / numa id" pair, which is more general.

protobuf/model_config.proto
```diff
@@ -239,6 +239,14 @@ message ModelInstanceGroup
   //@@     not be added to the scheduler. Default value is false.
   //@@
   bool passive = 7;
+
+  //@@  .. cpp:var:: string host_policy
+  //@@
+  //@@     The host policy name that the instance to be associated with.
+  //@@     The default value is set to reflect the device kind of the instance,
+  //@@     for instance, KIND_CPU is "cpu" and KIND_GPU is "gpu_<gpu_id>".
```
deadeyegoodwin, line 247: KIND_MODEL is "model"

deadeyegoodwin (changes requested):

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

PR #19, opened 2021-06-09
Fix segfault in Add

jbkyang-nvi:
Fixing https://github.com/triton-inference-server/server/issues/2893
and potential segfault issue in Append


include/triton/common/triton_json.h
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: remove "(c)" to avoid multi-line.

deadeyegoodwin (changes requested):
Also fix the title of the PR to say something about "JSON" so it is not so generic.

deadeyegoodwin (approved):

PR #23, opened 2021-08-09
Adding API to set/overwrite string

tanmayv25:
Needed for TRT autofiller logic

### DIFF
--- a/include/triton/common/triton_json.h
+++ b/include/triton/common/triton_json.h
@@ -209,6 +209,15 @@ class TritonJson {
       return TRITONJSON_STATUSSUCCESS;
     }
 
+    // Set/overwrite a string in a value. This changes the
+    // type of the value to string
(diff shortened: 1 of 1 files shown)


GuanLuo (approved):
We can add this for completeness of the TritonJson API. I am just curious that how it is going to be used in TRT autocompletion? We have autocompletion for ORT backend as well, but I don't think it encounters something similar.

tanmayv25:
It is being used here: https://github.com/triton-inference-server/tensorrt_backend/blob/tanmayv-dev/src/tensorrt.cc#L846

For TensorRT, we allow user to skip the data_type field in Inputs. The conversion sets such fields to 'TYPE_INVALID'. We will overwrite the `TYPE_INVALID` to the data type read from the TRT engine.

CoderHam (approved):

PR #26, opened 2021-09-20
add more configuration params for ORT

askhade:
Enable more optimizations for ORT.

### DIFF
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -814,6 +814,22 @@ message ModelOptimizationPolicy
     //@@       Currently only recognized by TensorRT backend.
     //@@
     bool output_copy_stream = 4;
+
+    //@@    .. cpp:var:: map<string, string> parameters
+    //@@
+    //@@       Additional paremeters used to configure cuda options.
+    //@@       Following parameters can be specified:
+    //@@         "cuda_conv_algo_search": EXHAUSTIVE (expensive exhaustive benchmarking 
+    //@@          using cudnnFindConvolutionForwardAlgorithmEx), HEURISTIC (lightweight 
+    //@@          heuristic based search using cudnnGetConvolutionForwardAlgorithm_v7)
+    //@@          Default is EXHAUSTIVE.
+    //@@         "gpu_mem_limit":  Default is std::numeric_limits<size_t>::max()
+    //@@         "arena_extend_strategy": Strategy used to grow the memory arena.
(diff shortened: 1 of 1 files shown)


protobuf/model_config.proto
```diff
@@ -814,6 +814,22 @@ message ModelOptimizationPolicy
     //@@       Currently only recognized by TensorRT backend.
     //@@
     bool output_copy_stream = 4;
+
```
deadeyegoodwin, line 817: Can you move these changes (and below as well) to be parameters similar to the existing ones in the ort backend. Putting backend-specific parameters in this file is something we are trying to get away from.

deadeyegoodwin (changes requested):

protobuf/model_config.proto
askhade: Do you mean these parameters in ORT backend?
```
optimization { execution_accelerators {
  gpu_execution_accelerator : [ {
    name : "tensorrt"
    parameters { key: "precision_mode" value: "FP16" }
    parameters { key: "max_workspace_size_bytes" value: "1073741824" }}]
}}
```

Well these are part of model_config.proto. I explicitly added a generic "map<string, string> parameters" so that these new fields are not runtime specific. Any runtime can use this field to add extra optimization configs.

askhade: I added "Currently only recognized by ONNXRuntime backend." because the example parameters in comments are only recognized by onnxruntime. Perhaps I can rewords the comments.

deadeyegoodwin: I realize that we have existing parameters that are marked as TensorRT only... and perhaps some of them could maybe be generalized for use in another backend... but that doesn't actually seem likely. So going forward we want backend parameters to use the existing generic parameters support, like here: https://github.com/triton-inference-server/tensorflow_backend#parameters

deadeyegoodwin: Unless it is something we know that is very general, like the tensorrt optimization parameters

askhade: Ok got it - You are referring to : https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto#L1772 right?

Since these are all optimizations, I would like to add a generic parameters map to ModelOptimizationPolicy message. I can move the detailed comments out so that each backend can add the parameters they support in their repo documentation.

askhade:
Closing this as we decided to use the existing parameters map for such optimizations instead of adding new fields.

PR #28, opened 2021-09-27
Add state description to the protobuf

Tabrizian:


protobuf/model_config.proto
```diff
@@ -1260,6 +1260,37 @@ message ModelSequenceBatching
...
+  {
+    //@@    .. cpp:var:: string input_name
+    //@@
+    //@@       The name of the model input.
+    //@@
+    string input_name = 1;
+
+    //@@    .. cpp:var:: string output_name
+    //@@
+    //@@       The name of the model input.
```
CoderHam, line 1277: input -> output

CoderHam (changes requested):

protobuf/model_config.proto
```diff
@@ -1377,6 +1408,16 @@ message ModelSequenceBatching
...
   //@@
   repeated ControlInput control_input = 2;
+
+  //@@  .. cpp:var:: State state (repeated)
+  //@@
+  //@@     The implicit state that should be stored in Triton for performing
+  //@@     inference requests in a sequence. Each sequence holds an implicit
+  //@@     state local to itself. The output state provided by the model in 
+  //@@     'output_name' field of the current inference request will be sent 
+  //@@     to the next request in the sequence in the 'input_name' field.
```
GuanLuo, line 1418: What will be the state as input of the first request in the sequence?

protobuf/model_config.proto
```diff
@@ -1263,6 +1263,37 @@ message ModelSequenceBatching
     repeated Control control = 2;
   }
 
+  //@@  .. cpp:var:: message State
+  //@@
+  //@@     Description of the implicit state for each model.
```
deadeyegoodwin, line 1268: "An input / output pair of tensors that carry state for the sequence."

deadeyegoodwin (changes requested):

protobuf/model_config.proto
```diff
@@ -1377,6 +1408,16 @@ message ModelSequenceBatching
   //@@     model.
   //@@
   repeated ControlInput control_input = 2;
+
+  //@@  .. cpp:var:: State state (repeated)
+  //@@
+  //@@     The implicit state that should be stored in Triton for performing
```
tanmayv25, line 1414: From the language it appears we require the users to use implicit state management for performing inference requests in a sequence. I don't think that is right.. We must explicitly call the feature is optional.

protobuf/model_config.proto
```diff
@@ -1377,6 +1408,16 @@ message ModelSequenceBatching
...
+  //@@     inference requests in a sequence. Each sequence holds an implicit
+  //@@     state local to itself. The output state provided by the model in
```
tanmayv25, line 1416: The language appears little confusing saying the link is between "output_name" and "input_name" fields. We must be more explicit saying the tensor contents get transferred and not just the "*_name" fields.

protobuf/model_config.proto
Tabrizian: The state will contain garbage data in the first inference request. I'll add a sentence about this.

protobuf/model_config.proto
Tabrizian: Will modify this to indicate that this is optional.

protobuf/model_config.proto
Tabrizian: I improved the wording. Let me know if it is still unclear.

CoderHam (dismissed):

tanmayv25 (approved):

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #30, opened 2021-10-07
Add response cache enable setting in model config

tanmayv25:


protobuf/model_config.proto
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: Remove '(c)' and fix line wrap

protobuf/model_config.proto
```diff
@@ -1661,6 +1662,16 @@ message ModelConfig
...
 
+  //@@  .. cpp::var:: bool response_cache_enable
+  //@@
+  //@@     Whether or not to use response cache for the model. If True, the
+  //@@     responses from the model are cached and when identical request
+  //@@     is encountered, instead of going through the model execution,
+  //@@     the response from the cache is utilized. By default, response
+  //@@     cache is disabled for the models.
+  //@@
+  bool response_cache_enable = 24;
```
deadeyegoodwin, line 1673: To future proof, can you add a ModelResponseCache message that has a single "bool enable" entry for now. Then after model_repository_agents add "ModelResponseCache response_cache"

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #35, opened 2021-12-07
Add optional field in ModelInput message

GuanLuo:
@deadeyegoodwin look and feel for the new field in model config

### DIFF
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -402,10 +402,8 @@ message ModelInput
   //@@  .. cpp:var:: bool optional
   //@@
   //@@     Whether or not the input is optional for the model execution.
-  //@@     If true, the input is not required in the inference request and
-  //@@     Triton will skip the input for dynamic batching check even if the
-  //@@     input is provided, as if 'allow_ragged_batch' is true. Default value
-  //@@     is false.
+  //@@     If true, the input is not required in the inference request.
+  //@@     Default value is false.
   //@@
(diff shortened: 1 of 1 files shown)


protobuf/model_config.proto
```diff
@@ -398,6 +398,16 @@ message ModelInput
   //@@     has a different shape in each request.
   //@@
   bool allow_ragged_batch = 7;
+
+  //@@  .. cpp:var:: bool optional
+  //@@
+  //@@     Whether or not the input is optional for the model execution.
+  //@@     If true, the input is not required in the inference request and
```
deadeyegoodwin, line 405: The input is optional... but it the input is actual present it should be treated just like any other input... so don't skip any dynamic batcher checks or other operations.

deadeyegoodwin (changes requested):

protobuf/model_config.proto
GuanLuo: I think when the input is optional, allowing ragged batching is the simple case as conceptually the request without the input has [0] shape. But in the case where ragged batching is not allowed and the input is optional, how should dynamic batcher react with a mix of requests with and without the input? Similar with the [0] shape and don't batch them?

deadeyegoodwin: Yes. A request can't be included in the batch if it doesn't have all the same inputs as the requests that are already in the batch.

deadeyegoodwin (approved):

PR #36, opened 2021-12-07
Add state initialization setting to model config protobuf

Tabrizian:
The setting is based on ModelWarmup. I have removed the parameters that where not related such as batch size and random_data.

### DIFF
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -1270,61 +1270,47 @@ message ModelSequenceBatching
   //@@
   message InitialState
   {
+    //@@      .. cpp:var:: DataType data_type
     //@@
-    //@@    .. cpp:var:: message StateMetaData
+    //@@         The data-type of the state.
     //@@
-    //@@       Meta data associated with a state.
+    DataType data_type = 1;
+
+    //@@      .. cpp:var:: int64 dims (repeated)
     //@@
-    message StateMetaData
+    //@@         The shape of the state tensor, not including the batch dimension.
+    //@@
+    repeated int64 dims = 2;
+
+    //@@      .. cpp:var:: oneof state_data
+    //@@
+    //@@         Specify how the initial state data is generated.
+    //@@
+    oneof state_data
     {
-      //@@      .. cpp:var:: DataType data_type
       //@@
-      //@@         The data-type of the state.
+      //@@      .. cpp:var:: bool zero_data
       //@@
-      DataType data_type = 1;
+      //@@         The identifier for using zeros as initial state data.
+      //@@         Note that the value of 'zero_data' will not be checked,
+      //@@         instead, zero data will be used as long as the field is set.
+      //@@
+      bool zero_data = 3;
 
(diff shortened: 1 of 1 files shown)


protobuf/model_config.proto
```diff
@@ -1578,6 +1584,70 @@ message ModelWarmup
   map<string, Input> inputs = 3;
 }
 
+//@@
```
tanmayv25, line 1587: I would suggest nesting it inside ModelSequenceBatching. Just before message State. Follow how Control is described for ControlInput.

protobuf/model_config.proto
```diff
@@ -1578,6 +1584,70 @@ message ModelWarmup
...
+  //@@
+  //@@     The name of the state initialization.
+  //@@
+  string name = 1;
+
+  //@@  .. cpp:var:: map<string, State> states
+  //@@
+  //@@     The state initialization meta data associated with every model state.
+  //@@
+  map<string, State> states = 2;
```
deadeyegoodwin, line 1648: What is this map for? Isn't this the state associated with a specific input/output pair? Do we need the nested State?

deadeyegoodwin (changes requested):

protobuf/model_config.proto
Tabrizian: Good catch. Fixed it in the next commit. We do not support nested states.

protobuf/model_config.proto
Tabrizian: Fixed

Tabrizian:
Below is an example Model Configuration that has state initialization description in it:

```protobuf
  state [
    {
      input_name: "INPUT_STATE"
      output_name: "OUTPUT_STATE"
      data_type: TYPE_INT32
      dims: 1
      initial_state: {
         state: {
            dims: 1
            data_type: TYPE_INT32
            zero_data: true
         }
         name: "state init"
      }
    }
  ]
```

protobuf/model_config.proto
```diff
@@ -1263,6 +1263,70 @@ message ModelSequenceBatching
...
+  //@@     Settings used to initialize data for implicit state.
+  //@@
+  message InitialState
+  {
+    //@@
+    //@@    .. cpp:var:: message StateMetaData
+    //@@
+    //@@       Meta data associated with a state.
+    //@@
+    message StateMetaData
```
deadeyegoodwin, line 1278: Why is this nested message needed? Can't data_type, dims and state_data be directly in InitialState?

deadeyegoodwin (changes requested):

protobuf/model_config.proto
deadeyegoodwin: Yes, I would keep datatype so that the state description is complete in itself. We should verify that the state datatype matches the tensor datatype.

deadeyegoodwin (approved):

tanmayv25 (approved):

GuanLuo (approved):

PR #38, opened 2022-01-27
WIP: Add support for I/O tensor structure to model config

CoderHam:
- Will add descriptions once finalized

Reasons for revised protobuf design:
1. map cannot be a repeated field
2. oneof cannot contain repeated fields

### DIFF
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -1739,24 +1739,19 @@ message TensorStructure
 {
   message ListStructure
   {
-    repeated TensorStructure list = 1;
+    repeated TensorStructure item = 1;
   }
 
-  message TupleStructure
+  message ReadOnlyListStructure
   {
-    repeated TensorStructure tuple = 1;
+    repeated TensorStructure item = 1;
   }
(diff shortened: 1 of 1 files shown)


protobuf/model_config.proto
```diff
@@ -1734,6 +1734,30 @@ message ModelResponseCache
   bool enable = 1;
 }
 
+message ListStruct
```
deadeyegoodwin, line 1737: Nest these 3 messages inside of TensorStructure, and we should be consistent and use Structure instead of Struct

deadeyegoodwin (changes requested):

protobuf/model_config.proto
```diff
@@ -1734,6 +1734,37 @@ message ModelResponseCache
   bool enable = 1;
 }
 
+
```
tanmayv25, line 1737: Should add the documentation on these fields as well.

CoderHam: Yes, I will add them once I have finalize the structure

PR #39, opened 2022-01-31
Add support for nan/inf in parsing/writing with rapidjson

rmccorm4:


deadeyegoodwin (approved):

rmccorm4:
Can't pass `writeFlags` to `rapidjson::prettyWriter` with latest release v1.1.0 of rapidjson-dev, the issue is as described here: https://github.com/Tencent/rapidjson/issues/905#issuecomment-370981353

We would need to build rapidjson from source to capture the latest fixes.

However, `PrettyWrite` is only used for displaying model configs internally currently, so this should not be an issue.

deadeyegoodwin (approved):

PR #40, opened 2022-02-02
Add GRPC trace service

GuanLuo:


protobuf/grpc_service.proto
```diff
@@ -1477,3 +1486,74 @@ message CudaSharedMemoryUnregisterRequest
...
+
+  //@@  .. cpp:var:: map<string,SettingValue> settings
+  //@@
+  //@@     The new setting values to be updated,
+  //@@     settings that are not specified will remain unchanged.
+  //@@
+  map<string, SettingValue> settings = 1;
+
+  //@@
+  //@@  .. cpp:var:: string name
```
deadeyegoodwin, line 1520: model_name

protobuf/grpc_service.proto
```diff
@@ -1477,3 +1486,74 @@ message CudaSharedMemoryUnregisterRequest
...
+  string model_name = 2;
+
+  //@@
+  //@@  .. cpp:var:: bool clear_setting
+  //@@
+  //@@     Clear the current model specific setting if specified, and
+  //@@     any new setting values will not be applied.
+  //@@     This option will be ignored if 'model_name' is not given.
+  //@@
+  bool clear_setting = 3;
```
deadeyegoodwin, line 1534: Why is this needed. Can we instead just say that if SettingValue:value is empty then the setting is cleared. If user doesn't want to change a value then they can just not provide it in 'settings'.

protobuf/grpc_service.proto
```diff
@@ -1477,3 +1486,74 @@ message CudaSharedMemoryUnregisterRequest
...
+    //@@    .. cpp:var:: string value (repeated)
+    //@@
+    //@@       The value.
+    //@@
+    repeated string value = 3;
+  }
+
+  //@@  .. cpp:var:: map<string,SettingValue> settings
+  //@@
+  //@@     The latest trace settings used.
```
deadeyegoodwin, line 1556: The current trace values, including any changes specified by TraceRequest.

deadeyegoodwin (changes requested):

protobuf/grpc_service.proto
GuanLuo: Different from HTTP where we can use GET / POST to separate whether the setting will be changed, in GRPC, an empty `settings` is ambiguous on whether the client want to clear the settings or to only retrieve the current setting.

I need to check on whether map has `has_xxx` API in proto3, if so, we may utilize that to only look up values if `has_settings == false` and a clear will be done if `has_settings == true && settings.size() == 0`

deadeyegoodwin: I think it is ok for GRPC to act like HTTP in that the response should always return all the global or model settings, no matter what (if anything) is specified in the request. Then anything specified in the request is considered an update to the setting.

tanmayv25, line 1534: clear_settings?

protobuf/grpc_service.proto
```diff
@@ -1477,3 +1486,74 @@ message CudaSharedMemoryUnregisterRequest
...
+  //@@  .. cpp:var:: bool clear_setting
+  //@@
+  //@@     Clear the current model specific setting if specified, and
+  //@@     any new setting values will not be applied.
+  //@@     This option will be ignored if 'model_name' is not given.
+  //@@
+  bool clear_setting = 3;
+}
+
+message TraceResponse
```
tanmayv25, line 1537: Which settings are returned if the request didn't set a model_name? I guess because the setting is applied globally same set of settings will be returned?
Shouldn't  there a way from client side just to retrieve trace settings for all models without updating them?

tanmayv25 (changes requested):

protobuf/grpc_service.proto
GuanLuo: If model name is not specified, global setting will be returned. I haven't considered the case where the user wants to get the setting for all models, so it is just either returning the global or model.

If the field is not specified in the request, then it won't update the field. So user can just send a request without `settings` to only retrieve the setting

protobuf/grpc_service.proto
```diff
@@ -202,6 +202,15 @@ service GRPCInferenceService
       returns (CudaSharedMemoryUnregisterResponse)
   {
   }
+
+  //@@  .. cpp:var:: rpc Trace(TraceRequest)
+  //@@                   returns (TraceResponse)
+  //@@
+  //@@     Update and get the trace setting of the Triton server.
+  //@@
+  rpc Trace(TraceRequest) returns (TraceResponse)
```
CoderHam, line 211: This should be renamed to TraceSettings since it does not get the trace but the settings for the trace

protobuf/grpc_service.proto
```diff
@@ -1477,3 +1486,74 @@ message CudaSharedMemoryUnregisterRequest
 //@@   Response message for CudaSharedMemoryUnregister.
 //@@
 message CudaSharedMemoryUnregisterResponse {}
+
+//@@
+//@@.. cpp:var:: message TraceRequest
+//@@
+//@@   Request message for Trace.
+//@@
+message TraceRequest
```
CoderHam, line 1495: TraceSettingsRequest

protobuf/grpc_service.proto
```diff
@@ -1477,3 +1486,74 @@ message CudaSharedMemoryUnregisterRequest
...
+  //@@     The values to be associated with a trace setting.
+  //@@
+  message SettingValue
+  {
+    //@@
+    //@@    .. cpp:var:: string value (repeated)
+    //@@
+    //@@       The value.
+    //@@
+    repeated string value = 3;
```
CoderHam, line 1509: Is there a reason for using 3 instead of 1?

protobuf/grpc_service.proto
CoderHam, line 1537: TraceSettingsResponse

protobuf/grpc_service.proto
```diff
@@ -1477,3 +1486,74 @@ message CudaSharedMemoryUnregisterRequest
...
```
CoderHam, line 1551: Use 1 instead of 3

CoderHam (changes requested):

protobuf/grpc_service.proto
deadeyegoodwin: That's a good point, in the future if we want to add other trace API we will be in trouble so we should be more specific here. TraceSettings is a good choice.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

CoderHam (approved):

tanmayv25 (approved):

PR #43, opened 2022-02-25
Add batch input item shape specification

GuanLuo:


protobuf/model_config.proto
```diff
@@ -509,6 +509,28 @@ message BatchInput
...
+    //@@         with shape [2, 2] are batched, the batch input will
+    //@@         have shape [3, 2] and value [ [3, 1], [3, 1], [2, 2]].
+    //@@
+    BATCH_ITEM_SHAPE = 4;
+
+    //@@      .. cpp:enumerator:: Kind::BATCH_ITEM_SHAPE_FLATTEN = 5
+    //@@
+    //@@         Among the requests in the batch, the shape of the
+    //@@         'source_input' will be added as input with shape
+    //@@         [batch_size * len(input_dim)]. For example, if one
```
CoderHam, line 528: Fix to mention flat structure of shape

CoderHam (changes requested):
Both descriptions for BATCH_ITEM_SHAPE and BATCH_ITEM_SHAPE_FLATTEN are identical except in the flatten case the values with be flat. Need to clarify that the values in the second case are flattened

protobuf/model_config.proto
```diff
@@ -509,6 +509,28 @@ message BatchInput
     //@@         The data of the tensor will be uninitialized.
```
CoderHam, line 509: Update copyright?

CoderHam (changes requested):

protobuf/model_config.proto
GuanLuo: Can you clarify this comment?

tanmayv25: I think we must explicitly call out single dimensionality of tensor. It is hard to ascertain that with just [batch_size * len(input_dim)]

CoderHam (approved):

PR #50, opened 2022-05-03
Add Functions to TritonJson API

nv-kmcgill53:
Adds two new functions to the TritonJson api for use in the tensorflow backend.

### DIFF
--- a/include/triton/common/triton_json.h
+++ b/include/triton/common/triton_json.h
@@ -575,6 +575,23 @@ class TritonJson {
       return TRITONJSON_STATUSSUCCESS;
     }
 
+    // Remove member from this object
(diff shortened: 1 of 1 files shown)


include/triton/common/triton_json.h
```diff
@@ -243,6 +243,29 @@ class TritonJson {
       return TRITONJSON_STATUSSUCCESS;
     }
 
+    // Set/overwrite a string member, name, in this object with provided value
```
tanmayv25, line 246: Should we improve the documentation like below?

Set/overwrite a string member with provided name and value in this object ?

nv-kmcgill53: updated.

tanmayv25 (approved):

PR #51, opened 2022-05-05
Extend GRPC ModelRepositoryParameter to allow bytes

GuanLuo:


protobuf/grpc_service.proto
```diff
@@ -1142,6 +1142,12 @@ message ModelRepositoryParameter
     //@@       A string parameter value.
     //@@
     string string_param = 3;
+
```
tanmayv25, line 1145: You must also add this to kfserve copy here: https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/grpc_predict_v2.proto#L276

GuanLuo: I don't think it is needed, the kfserve one only covers the basic protocol (infer), whereas this is the Triton specific extension. The protobuf message is added within Triton domain

rmccorm4: I agree with Guan. If this was `InferParameter`, it should be updated in KServer. But `ModelRepositoryParameter` is Triton-specific

rmccorm4 (approved):

protobuf/grpc_service.proto
tanmayv25: Ahh.. That's correct. Missed that point.

tanmayv25 (approved):

PR #52, opened 2022-05-09
Add lock for logging

krishung5:
Using the server.log from L0_long_running_stress test for testing.
Before adding the log, there are some overlapped loggings:
image

After adding the log, there is no overlapped logging:
image


include/triton/common/logging.h
```diff
@@ -76,36 +77,34 @@ class Logger {
  private:
   std::vector<bool> enables_;
   uint32_t vlevel_;
+  static std::mutex mutex;
```
rmccorm4, line 80: Use underscore for member variables: `mutex_`

rmccorm4 (changes requested):

rmccorm4:
@GuanLuo does `LogMessage()` need similar locking? Or is `Log()` the main interface used internally?

src/logging.cc
```diff
@@ -46,10 +46,14 @@ Logger gLogger_;
...
 Logger::Logger() : enables_{true, true, true}, vlevel_(0) {}
 
+std::mutex Logger::mutex;
+
 void
 Logger::Log(const std::string& msg)
 {
+  mutex.lock();
   std::cerr << msg << std::endl;
+  mutex.unlock();
```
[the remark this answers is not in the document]
krishung5: Thanks for the link! Used `std::lock_guard<std::mutex>` instead.

src/logging.cc
```diff
@@ -46,9 +46,12 @@ Logger gLogger_;
 
 Logger::Logger() : enables_{true, true, true}, vlevel_(0) {}
 
+std::mutex Logger::mutex;
```
tanmayv25, line 49: mutex -> mutex_

krishung5: Thanks for the catch

tanmayv25:
> @GuanLuo does `LogMessage()` need similar locking? Or is `Log()` the main interface used internally?

I don't think LogMessage needs the locking. Each LOG_INFO or similar call creates a separate LogMessage object, which holds the message in a std::stringstream object. Upon the destruction of this object, it  Logger::Log to place the string on std::cerr. So the mutex is at the right place.

include/triton/common/logging.h
```diff
@@ -76,36 +77,34 @@ class Logger {
  private:
   std::vector<bool> enables_;
   uint32_t vlevel_;
+  static std::mutex mutex_;
```
GuanLuo, line 80: Why `static`? We have only one global `Logger` object where all writes go through it, so local member variable should be sufficient.

krishung5:
> > @GuanLuo does `LogMessage()` need similar locking? Or is `Log()` the main interface used internally?
> 
> I don't think LogMessage needs the locking. Each LOG_INFO or similar call creates a separate LogMessage object, which holds the message in a std::stringstream object. Upon the destruction of this object, it Logger::Log to place the string on std::cerr. So the mutex is at the right place.

I was having the same thought, but after printing some number at the beginning of the logging in Logger::Log function:

`9314---I0509 21:16:59.191318 24600 grpc_server.cc:3971] New request handler for ModelStreamInferHandler, 0
9315---I0509 21:16:59.209009 24600 grpc_server.cc:3978] Process for ModelStreamInferHandler, rpc_ok=0, context 0, 0 step READ
9316---I0509 21:16:59.209206 24600 grpc_server.cc:3978] Process for ModelStreamInferHandler, rpc_ok=1, context 0, 0 step COMPLETE
9317---I0509 21:16:59.209261 24600 grpc_server.cc:2504] Done for ModelStreamInferHandler, 0
**I0509 21:16:59.220848 24600 identity.cc:843] request 0: id = "", correlation_id = 0, input_count = 1, requested_output_count = 1
I0509 21:16:59.220925 24600 identity.cc:1035] 	input INPUT0: datatype = FP32, shape = [1,1], byte_size = 4, buffer_count = 1
I0509 21:16:59.220946 24600 identity.cc:1044] 	requested_output OUTPUT0
I0509 21:16:59.220967 24600 infer_response.cc:166] add response output: output: OUTPUT0, type: FP32, shape: [1,1]**
9318---I0509 21:16:59.220997 24600 grpc_server.cc:2594] GRPC: using buffer for 'OUTPUT0', size: 4, addr: 0x7f540c0050a0`

I found that some of the loggings are not running through the Logger::Log function. Is there any other place where we should also place the lock?

include/triton/common/logging.h
krishung5: Updated, please review.

GuanLuo:
The backend will use the [logging API](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonserver.h#L267-L276) if they want to use Triton's logging, you can trace that to see what is going on

GuanLuo (commented):
LGTM, but still need to resolve the backend logging inconsistency?

krishung5:
@GuanLuo Sorry for the late reply. I think there was some compiled issue. I re-run the build and the devel container and the logging works fine.

GuanLuo (approved):

tanmayv25 (approved):

rmccorm4 (approved):

PR #48, opened 2022-04-15
Move model_config to common

kpedro88:
Addresses https://github.com/triton-inference-server/server/issues/2759: many of the functions in `model_config.h` are generically useful even on the client side and only depend on the generated protobuf headers. Therefore, it is useful to have them more broadly accessible via the common repo.

The header and cc files are copied here from the core repo (except for a few functions with non-trivial dependencies that stay in core), with the namespace changed to `triton::common`. A new library `triton-common-model-config` is built. Because many of the functions depend on the generated protobuf headers, this library is only built when the protobuf part of this repo is enabled.

This PR goes with https://github.com/triton-inference-server/core/pull/60.

PR #54, opened 2022-05-12
Fix missing BF16 handle

GuanLuo:


mc-nv (approved):

src/model_config.cc
```diff
@@ -414,12 +418,22 @@ ProtocolStringToDataType(const char* dtype, size_t len)
       return inference::DataType::TYPE_FP64;
     }
   } else if (*dtype == 'B') {
-    if (dtype[1] == 'Y') {
-      if (!strcmp(dtype + 2, "TES")) {
-        return inference::DataType::TYPE_STRING;
-      }
-    } else if (!strcmp(dtype + 1, "OOL")) {
-      return inference::DataType::TYPE_BOOL;
+    switch (dtype[1]) {
```
rmccorm4, line 421: @GuanLuo This file seems mostly duplicated from [core/src/model_config.cc](https://github.com/triton-inference-server/core/blob/32c944f84f8d2952f6631a34d1139c6ef93af519/src/model_config.cc) after this [PR](https://github.com/triton-inference-server/common/pull/48) - is this intended? or do we plan to remove the core version? I notice there are some functions not included in both.

rmccorm4 (changes requested):

src/model_config.cc
rmccorm4, line 421: @GuanLuo This file seems mostly duplicated from [core/src/model_config.cc](https://github.com/triton-inference-server/core/blob/32c944f84f8d2952f6631a34d1139c6ef93af519/src/model_config.cc) after this [PR](https://github.com/triton-inference-server/common/pull/48) - is this intended? or do we plan to remove the core version? I notice there are some functions not included in both.

rmccorm4 (commented):
@GuanLuo This file seems mostly duplicated from [core/src/model_config.cc](https://github.com/triton-inference-server/core/blob/32c944f84f8d2952f6631a34d1139c6ef93af519/src/model_config.cc) after this [PR](https://github.com/triton-inference-server/common/pull/48) - is this intended? or do we plan to remove the core version? I notice there are some functions not included in both.

src/model_config.cc
GuanLuo: The PR is for moving generic utils in `model_config.h/.cc` from core repo to common repo, the missing functions are core specific and moved to `model_config_utils.h/.cc`. There is still a copy `model_config.cc` in core but it seems to be leftover from the move, it is not being used anywhere

rmccorm4: Filed DLIS-3781 to follow-up. LGTM

PR #57, opened 2022-05-23
Add different logging format

GuanLuo:
Default
```
I0523 23:49:06.674147 141 server.cc:288] All models are stopped, unloading models
```
ISO8601
```
2022-05-23T23:49:14 I 153 server.cc:288] All models are stopped, unloading models
```

### DIFF
--- a/include/triton/common/logging.h
+++ b/include/triton/common/logging.h
@@ -103,6 +103,8 @@ extern Logger gLogger_;
 #define LOG_SET_VERBOSE(L)                  \
   triton::common::gLogger_.SetVerboseLevel( \
       static_cast<uint32_t>(std::max(0, (L))))
+#define LOG_SET_FORMAT(F) \
+  triton::common::gLogger_.SetLogFormat((F))
 
 #ifdef TRITON_ENABLE_LOGGING
 
--- a/include/triton/common/logging.h
+++ b/include/triton/common/logging.h
@@ -51,6 +51,8 @@ class LogMessage {
 // Global logger for messages. Controls how log messages are reported.
 class Logger {
  public:
+  enum Format { kDEFAULT = 0, kISO8601 = 1 };
+
   Logger();
 
   // Is a log level enabled.
@@ -68,6 +70,12 @@ class Logger {
   // Set the current verbose logging level.
   void SetVerboseLevel(uint32_t vlevel) { vlevel_ = vlevel; }
 
+  // Get the logging format.
+  Format LogFormat() { return format_; }
+
+  // Set the logging format.
+  void SetLogFormat(Format format) { format_ = format; }
+
   // Log a message.
   void Log(const std::string& msg);
 
@@ -77,6 +85,7 @@ class Logger {
  private:
   std::vector<bool> enables_;
   uint32_t vlevel_;
+  Format format_;
   std::mutex mutex_;
 };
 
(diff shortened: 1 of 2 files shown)


dyastremsky (approved):

krishung5 (approved):

include/triton/common/logging.h
```diff
@@ -51,6 +51,8 @@ class LogMessage {
 // Global logger for messages. Controls how log messages are reported.
 class Logger {
  public:
+  enum Format { kDEFAULT = 0, kISO8601 = 1 };
```
rmccorm4, line 54: `enum class` instead? https://stackoverflow.com/a/18335862

tanmayv25:
Not required, but should we add a time zone designator in the timestamp? Looks like ISO 8601 goes about at length regarding the same. I would suggest `Local time with an offset to UTC`. 
It will be helpful for people to correlate logs from across the timezones.

src/logging.cc
```diff
@@ -70,30 +70,67 @@ LogMessage::LogMessage(const char* file, int line, uint32_t level)
...
+            << std::setw(2) << tm_time.tm_mday << " " << std::setw(2)
+            << tm_time.tm_hour << ':' << std::setw(2) << tm_time.tm_min << ':'
+            << std::setw(2) << tm_time.tm_sec << "." << std::setw(6) << tv.tv_usec
+            << ' ' << static_cast<uint32_t>(getpid()) << ' ' << path << ':'
+            << line << "] ";
 #endif
+    break;
+  }
+  case Logger::Format::kISO8601: {
+    // YYYY-MM-DDThh:mm:ss L
```
rmccorm4, line 105: Maybe an addition of a `Z` specifier at the end to denote this is in UTC would be helpful, similar to Tanmay's comment.

> As format Z,  This date time has no offset from UTC (or 0 hour offset).

src/logging.cc
```diff
@@ -70,30 +70,67 @@ LogMessage::LogMessage(const char* file, int line, uint32_t level)
     path = path.substr(pos + 1, std::string::npos);
   }
 
+  // 'L' for log level
```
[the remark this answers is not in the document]
rmccorm4: I believe it's saying the 'L' references in the following formats means Level:
i.e.
```
  case Logger::Format::kDEFAULT: {
    // LMMDD hh:mm:ss.ssssss     <----
...
  }
  case Logger::Format::kISO8601: {
    // YYYY-MM-DDThh:mm:ss L     <----
...
```

src/logging.cc
GuanLuo: @tanmayv25 I think I will use `Z` to highlight UTC time as described here. The doc for `gettimeofday` suggests that there can be complication when handling local time.
> The use of the timezone structure is obsolete; the tz argument
       should normally be specified as NULL.  (See NOTES below.)

src/logging.cc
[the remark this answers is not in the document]
dyastremsky: I see, thanks for clarifying!

rmccorm4 (approved):

krishung5 (approved):

tanmayv25 (approved):

tanmayv25 (approved):

dyastremsky (approved):

PR #60, opened 2022-06-08
Add ThreadPool to common

rmccorm4:
Add simple ThreadPool implementation that doesn't rely on `boost::asio`

[common] Adds custom ThreadPool implementation in common: https://github.com/triton-inference-server/common/pull/60
[core] Use thread pool for loading models in model_repository_manager.cc: https://github.com/triton-inference-server/core/pull/86
[server] Expose num threads to server CLI: https://github.com/triton-inference-server/server/pull/4500

> NOTE: Will update `AsyncWorkQueue` to use `ThreadPool` internally in a follow-up ticket (DLIS-3860) to only maintain one implementation, and not have `ThreadPool` be restricted to only being a singleton like AsyncWorkQueue currently is.

### DIFF
--- a/include/triton/common/thread_pool.h
+++ b/include/triton/common/thread_pool.h
@@ -44,7 +44,7 @@ class ThreadPool {
   using Task = std::function<void(void)>;
   // Assigns "task" to the task queue for a worker thread to execute when
   // available. This will not track the return value of the task.
-  void enqueue(Task task);
+  void enqueue(Task&& task);
 
  private:
   std::queue<Task> task_queue_;
--- a/include/triton/common/thread_pool.h
+++ b/include/triton/common/thread_pool.h
@@ -28,6 +28,7 @@
 #include <condition_variable>
 #include <functional>
 #include <queue>
+#include <stdexcept>
 #include <thread>
 
 namespace triton { namespace common {
--- a/include/triton/common/thread_pool.h
+++ b/include/triton/common/thread_pool.h
@@ -28,7 +28,6 @@
 #include <condition_variable>
 #include <functional>
 #include <queue>
-#include <stdexcept>
 #include <thread>
 
 namespace triton { namespace common {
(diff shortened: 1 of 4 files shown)


src/thread_pool.cc
```diff
@@ -0,0 +1,96 @@
...
+  {
+    std::lock_guard<std::mutex> lk(queue_mtx_);
+    // Don't accept more work if pool is shutting down
+    if (stop_) {
+      return;
+    }
+    task_queue_.push(std::move(task));
+  }
+  // Only wake one thread per task
+  cv_.notify_one();
```
nv-kmcgill53, line 93: This isn't really going to mean much in the context of model loading, since we assume most of the calls are going to be all the `Load` calls are going to be all at once, but I can't find in the docs for `notify_one()` whether this multiplexes the threads/notifies next available/randomly available. As a performance enhancement we might put the threads on a stack so that they might stay "hot." This can be investigated in another ticket if we use this thread pool for anything closer to this use case.

rmccorm4: @nv-kmcgill53 can you file a separate ticket and link this code/comment? I agree it probably won't add much for the current context of model loading, but could be more impactful if used in other places in the future.

nv-kmcgill53: Ticket created: DLIS-3859

nv-kmcgill53 (dismissed):

include/triton/common/thread_pool.h
```diff
@@ -0,0 +1,59 @@
...
+ public:
+  explicit ThreadPool(std::size_t thread_count);
+  ~ThreadPool();
+  ThreadPool(const ThreadPool&) = delete;
+  ThreadPool& operator=(const ThreadPool&) = delete;
+
+  using Task = std::function<void(void)>;
+  // Assigns "task" to the task queue for a worker thread to execute when
+  // available. This will not track the return value of the task.
+  void enqueue(Task task);
```
GuanLuo, line 47: `Task&&`? Passing by value is going to make a new copy of the task

rmccorm4: Good catch

GuanLuo (approved):

tanmayv25 (approved):

PR #61, opened 2022-06-08
Add 'count' field for warmup

GuanLuo:


tanmayv25 (dismissed):

protobuf/model_config.proto
```diff
@@ -1665,6 +1665,15 @@ message ModelWarmup
   //@@     control tensors.
   //@@
   map<string, Input> inputs = 3;
+
+  //@@  .. cpp:var:: uint32 repeat_count
+  //@@
+  //@@     The number of iterations that this warmup sample will be repeated.
+  //@@     For example, if this field is set to 2, 2 model executions using this
```
tanmayv25, line 1672: Add a note about sequence model limitation here?

protobuf/model_config.proto
```diff
@@ -1665,6 +1665,20 @@ message ModelWarmup
...
+  //@@     sample will be scheduled for warmup. Default value is 0 which
```
rmccorm4, line 1673: Looking at core PR, I think the default value is 1, not 0? Also if user sets repeat_count to 0 or 1, they'll both have the same behavior right? Maybe make that more clear here if possible.

rmccorm4 (changes requested):

protobuf/model_config.proto
GuanLuo: Any suggestion on the wording? The "default value..." here is trying to clarify the behavior for 0 (protobuf default). When value is 1, I am thinking that the reader will relate the behavior to the "2" example above as they are non-zero value.

rmccorm4: Yeah repeat_count of 1 makes it sound like the request will be "repeated once", meaning 2 requests. But it looks like the code behavior would be one request for both repeat_count==0 or 1. 

Maybe `repeat_count` could be renamed to `count` to be less confusing if that's how the code is treating it? Simply the `count` of requests of this sample.

rmccorm4 (approved):

PR #63, opened 2022-06-15
Use thread_pool in async_work_queue internally

rmccorm4:
Backend: https://github.com/triton-inference-server/backend/pull/62
Common: https://github.com/triton-inference-server/common/pull/63
Core: https://github.com/triton-inference-server/core/pull/89

### DIFF
--- a/include/triton/common/thread_pool.h
+++ b/include/triton/common/thread_pool.h
@@ -44,9 +44,9 @@ class ThreadPool {
   using Task = std::function<void(void)>;
   // Assigns "task" to the task queue for a worker thread to execute when
   // available. This will not track the return value of the task.
-  void enqueue(Task&& task);
+  void Enqueue(Task&& task);
   // Returns the number of threads in thread pool
-  size_t size() { return workers_.size(); }
+  size_t Size() { return workers_.size(); }
 
  private:
   std::queue<Task> task_queue_;
--- a/include/triton/common/thread_pool.h
+++ b/include/triton/common/thread_pool.h
@@ -45,6 +45,8 @@ class ThreadPool {
   // Assigns "task" to the task queue for a worker thread to execute when
   // available. This will not track the return value of the task.
   void enqueue(Task&& task);
+  // Returns the number of threads in thread pool
+  size_t size() { return workers_.size(); }
 
  private:
   std::queue<Task> task_queue_;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,6 +141,7 @@ add_library(
   triton-common-async-work-queue
   src/async_work_queue.cc
   src/error.cc
+  src/thread_pool.cc
 )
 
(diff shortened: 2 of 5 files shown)


nv-kmcgill53:
What happens when two different threads try to use this AsyncWorkQueue? Both threads will need to know that the pool is initialized which means each thread calling `triton::common::AsyncWorkQueue::Initialize(n)`. So the second one is going to lose. Should we be able to resize the pool if the second initialize call has size greater than the first initialize call? This is tricky as the winning thread can start submitting work to the queue while the other one is asking for the resize. Maybe this is a known limitation and another ticket for improvement if the need arises. What do you think @tanmayv25 and @GuanLuo?

include/triton/common/thread_pool.h
```diff
@@ -45,6 +45,8 @@ class ThreadPool {
   // Assigns "task" to the task queue for a worker thread to execute when
   // available. This will not track the return value of the task.
   void enqueue(Task&& task);
+  // Returns the number of threads in thread pool
+  size_t size() { return workers_.size(); }
```
GuanLuo, line 49: Capitalize function names

GuanLuo:
I wonder if the `AsyncWorkQueue` even is initialized and used properly because nowadays the backends are compiled separately as shared library. `AsyncWorkQueue` was added when backends were compiled in core and thus the instance initialized during `InferenceServer::Init()` would be used by the backends. If this is no longer the case and the backend sees its own copy of `AsyncWorkQueue`, we should consider either move this utility completely to backend and let individual backend decide if it will use this feature (via backend config) or move this to core and expose it to backend via backend API.

Ref: https://stackoverflow.com/questions/8623657/multiple-instances-of-singleton-across-shared-libraries-on-linux

include/triton/common/thread_pool.h
rmccorm4: done

rmccorm4:
> I wonder if the `AsyncWorkQueue` even is initialized and used properly because nowadays the backends are compiled separately as shared library. `AsyncWorkQueue` was added when backends were compiled in core and thus the instance initialized during `InferenceServer::Init()` would be used by the backends. If this is no longer the case and the backend sees its own copy of `AsyncWorkQueue`, we should consider either move this utility completely to backend and let individual backend decide if it will use this feature (via backend config) or move this to core and expose it to backend via backend API.
> 
> Ref: https://stackoverflow.com/questions/8623657/multiple-instances-of-singleton-across-shared-libraries-on-linux

@GuanLuo do we have a ticket for this?

rmccorm4:
@nv-kmcgill53 

> What happens when two different threads try to use this AsyncWorkQueue? Both threads will need to know that the pool is initialized which means each thread calling triton::common::AsyncWorkQueue::Initialize(n). So the second one is going to lose.

I believe this is the intention given the singleton design of this class and existing unit test for this class, unless I'm misunderstanding the question

> Should we be able to resize the pool if the second initialize call has size greater than the first initialize call? This is tricky as the winning thread can start submitting work to the queue while the other one is asking for the resize.

For the AsyncWorkQueue singleton's current use, it's expected the queue would be destructed first (and cleaned up waiting for threads to finish), and then recreated with a different size. 

For the ThreadPool class, it's currently implemented as a fixed-size for simplicity, and can be improved to be more dynamic if the need arises in the future, but I didn't want to prematurely add and test that.

GuanLuo:
There is no ticket tracking my concern, it came up when I was reviewing this PR. I think the first step is to check if my concern is valid, i.e. if `--buffer-manager-thread-count` actually takes affect within backend execution (printing `AsyncWorkQueue::WorkerCount()` somewhere inside a backend)

GuanLuo (approved):

nv-kmcgill53 (approved):

PR #66, opened 2022-07-07
Incorporating Dynamic Logging

fpetrini15:


include/triton/common/logging.h
```diff
@@ -73,9 +72,25 @@ class Logger {
...
+      format = "ISO8601";
+    }
+    return format;
+  }
+
   // Set the logging format.
   void SetLogFormat(Format format) { format_ = format; }
 
+  // Get the log output file name.
+  std::string GetLogOutFile() { return filename_; }
```
GuanLuo, line 89: `const std::string& LogOutputFile()`

include/triton/common/logging.h
```diff
@@ -73,9 +72,25 @@ class Logger {
...
+
+  // Set the log output file.
+  void SetLogOutFile(std::string filename) { filename_ = filename; }
```
GuanLuo, line 92: `const std::string&`

include/triton/common/logging.h
```diff
@@ -103,8 +119,12 @@ extern Logger gLogger_;
 #define LOG_SET_VERBOSE(L)                  \
   triton::common::gLogger_.SetVerboseLevel( \
       static_cast<uint32_t>(std::max(0, (L))))
-#define LOG_SET_FORMAT(F) \
-  triton::common::gLogger_.SetLogFormat((F))
+#define LOG_VERBOSE_LEVEL triton::common::gLogger_.VerboseLevel()
```
GuanLuo, line 122: Move this down, group "getter" together

src/logging.cc
```diff
@@ -37,20 +37,40 @@
...
 #include <iostream>
 
 namespace triton { namespace common {
 
 Logger gLogger_;
 
-Logger::Logger() : enables_{true, true, true}, vlevel_(0), format_(Format::kDEFAULT) {}
+Logger::Logger()
+    : enables_{true, true, true}, vlevel_(0), format_(Format::kDEFAULT),
+      filename_("")
```
GuanLuo, line 50: Don't need it, std::string default initialize to empty string

src/logging.cc
```diff
@@ -37,20 +37,40 @@
...
 Logger::Log(const std::string& msg)
 {
   const std::lock_guard<std::mutex> lock(mutex_);
-  std::cerr << msg << std::endl;
+  if (Logger::GetLogOutFile() != "") {
+    try {
+      std::ofstream file_stream;
+      file_stream.open(filename_, std::ios::app);
+      file_stream << msg << std::endl;
+      file_stream.close();
```
GuanLuo, line 63: Any reason not to keep the file stream open until it is asked to log to different location?

src/logging.cc
```diff
@@ -37,20 +37,40 @@
...
+    }
+    catch (const std::ofstream::failure& e) {
+      std::cerr << "failed creating trace file: " << e.what();
+    }
+    catch (...) {
+      std::cerr << "failed creating trace file: reason unknown";
```
GuanLuo, line 69: std::endl

src/logging.cc
```diff
@@ -37,20 +37,40 @@
...
-  std::cerr << msg << std::endl;
+  if (Logger::GetLogOutFile() != "") {
+    try {
+      std::ofstream file_stream;
+      file_stream.open(filename_, std::ios::app);
+      file_stream << msg << std::endl;
+      file_stream.close();
+    }
+    catch (const std::ofstream::failure& e) {
+      std::cerr << "failed creating trace file: " << e.what();
```
GuanLuo, line 66: same as below

src/logging.cc
GuanLuo: >suggesting a new member variable that holds the stream open once invoked until the out file has been changed

Yes.

include/triton/common/logging.h
```diff
@@ -73,9 +73,25 @@ class Logger {
...
+  }
+
   // Set the logging format.
   void SetLogFormat(Format format) { format_ = format; }
 
+  // Get the log output file name.
+  const std::string& GetLogOutFile() { return filename_; }
+
+  // Set the log output file.
+  void SetLogOutFile(const std::string& filename) { filename_ = filename; file_name_changed_=true; }
```
Tabrizian, line 93: Is the code formatted? This line should be broken down to multiple lines.

src/logging.cc
```diff
@@ -44,13 +44,38 @@ namespace triton { namespace common {
...
+      }
+      else if(file_stream_.is_open() && file_name_changed_) {
+        file_stream_.close();
+        file_name_changed_ = false;
+        file_stream_.open(filename_, std::ios::app);
+      }
+      file_stream_ << msg << std::endl;
+    }
+    catch (const std::ofstream::failure& e) {
+      std::cerr << "failed creating trace file: " << e.what() << std::endl;
```
Tabrizian, line 71: Error message doesn't look correct.

include/triton/common/logging.h
```diff
@@ -73,9 +73,25 @@ class Logger {
...
+      format = "ISO8601";
+    }
+    return format;
+  }
+
   // Set the logging format.
   void SetLogFormat(Format format) { format_ = format; }
 
+  // Get the log output file name.
+  const std::string& GetLogOutFile() { return filename_; }
```
Tabrizian, line 90: We don't usually use `Get+` when retrieving attributes. Just say "LogFile".

include/triton/common/logging.h
Tabrizian: Do we need to lock the mutex here in case multiple clients are changing the logging simultaneously?

Tabrizian (changes requested):

include/triton/common/logging.h
fpetrini15: I can add a lock here, however, does that mean we should be adding a lock for every function where we change attributes?

include/triton/common/logging.h
```diff
@@ -73,9 +73,30 @@ class Logger {
...
+  // Get the log output file name.
+  const std::string& LogFile() { return filename_; }
+
+  // Set the log output file.
+  void SetLogFile(const std::string& filename)
+  {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    filename_ = filename;
+    file_name_changed_ = true;
+  }
```
GuanLuo, line 98: Why not just open the file stream here?

src/logging.cc
```diff
@@ -44,13 +44,37 @@ namespace triton { namespace common {
...
+        file_stream_.open(filename_, std::ios::app);
+      } else if (file_stream_.is_open() && file_name_changed_) {
+        file_stream_.close();
+        file_name_changed_ = false;
+        file_stream_.open(filename_, std::ios::app);
+      }
+      file_stream_ << msg << std::endl;
+    }
+    catch (const std::ofstream::failure& e) {
+      std::cerr << "failed creating log file: " << e.what() << std::endl;
```
GuanLuo, line 70: failed to open

include/triton/common/logging.h
fpetrini15: That's a great idea! It has been incorporated in the latest commit.

include/triton/common/logging.h
```diff
@@ -73,9 +74,39 @@ class Logger {
...
   Format LogFormat() { return format_; }
 
+  // Get the logging format as a string.
+  std::string LogFormatString()
+  {
+    std::string format = "default";
+    if (format_ == Format::kISO8601) {
+      format = "ISO8601";
+    }
+    return format;
```
GuanLuo, line 84:
```suggestion
    switch (format__) {
    case Format::kISO8601:
      return "ISO8601";
    case Format::kDEFAULT:
      return "default"
    }
```

include/triton/common/logging.h
```diff
@@ -73,9 +74,42 @@ class Logger {
...
+  // Get the logging format as a string.
+  std::string LogFormatString()
+  {
+    switch (format_) {
+      case Format::kISO8601:
+        return "ISO8601";
+      case Format::kDEFAULT:
+        return "default";
+      default:
+        return "";
```
GuanLuo, line 86: Do we want to return something when format is with unexpected value? If not, raise exception

include/triton/common/logging.h
```diff
@@ -73,9 +74,42 @@ class Logger {
...
+  const std::string& LogFile() { return filename_; }
+
+  // Set the log output file.
+  bool SetLogFile(const std::string& filename)
+  {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    file_stream_.close();
+    filename_ = filename;
+    if(!filename.empty()) {
+      file_stream_.open(filename_, std::ios::app);
```
GuanLuo, line 103: Handle exception

GuanLuo (changes requested):

include/triton/common/logging.h
fpetrini15: It is not possible to get this condition with the current code. The values are filtered prior to calling this function. It was only included because the compiler complained. Will an "Invalid format" just in case future changes impact this assumption.

include/triton/common/logging.h
fpetrini15: Error is handled below. Will revert to the previous log file and return false. Return status will be handled on the server side.

GuanLuo (approved):

PR #67, opened 2022-07-25
Fix rapidjson memory leak issue

krishung5:
Using `rapidjson::Value(static_cast<rapidjson::Type>(type))` to create Value causes memory leak even using allocator to release the object. Changed to use `allocator_->Malloc()` to create Value and modified `Releae()` function accordingly.
Related PRs:
backend: https://github.com/triton-inference-server/backend/pull/67
tensorrt_backend: https://github.com/triton-inference-server/tensorrt_backend/pull/44
server: https://github.com/triton-inference-server/server/pull/4693

### DIFF
--- a/include/triton/common/triton_json.h
+++ b/include/triton/common/triton_json.h
@@ -36,7 +36,7 @@
 #endif
 #include <rapidjson/document.h>
 #pragma GCC diagnostic pop
-#endif // _WIN32
+#endif  // _WIN32
 
 #include <rapidjson/allocators.h>  // CrtAllocator (default) for Writer instantiation
 #include <rapidjson/encodings.h>  // UTF8 (default) for Writer instantiation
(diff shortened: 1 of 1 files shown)


tanmayv25 (approved):

include/triton/common/triton_json.h
```diff
@@ -124,9 +124,10 @@ class TritonJson {
 
     // Construct a non-top-level JSON value in a 'document'.
     explicit Value(TritonJson::Value& document, const ValueType type)
```
GuanLuo, line 126: When does this get called? I wonder if adding destructor to the class would solve the issue

krishung5: This constructor is usually used for auto-complete like [this](https://github.com/triton-inference-server/tensorflow_backend/blob/main/src/tensorflow.cc#L1235-L1237).
I tried adding destructor to the class either simply using delete to clean up the Value pointer or using allocator to release Value object. The first case, it will cause double free issue of Value when deleting the DOM object. The second case, the memory leak issue still exists.

rmccorm4: If the destructor calls `Release()` (which first checks `if value == nullptr` before delete) does this still have a double free issue?

krishung5: No, it won't have a double free issue but it will still have memory leak.

PR #73, opened 2022-08-15
Update documentation for execution accelerators

Tabrizian:


protobuf/model_config.proto
```diff
@@ -897,15 +897,16 @@ message ModelOptimizationPolicy
...
-    //@@       For ONNX Runtime backend, possible value is "tensorrt" as name,
-    //@@       and no parameters are required.
+    //@@       For ONNX Runtime backend, possible value is "tensorrt".
     //@@
     //@@       For TensorFlow backend, possible values are "tensorrt",
-    //@@       "auto_mixed_precision", "gpu_io".
+    //@@       "auto_mixed_precision", "gpu_io". Tensorflow backend does not
+    //@@       support "INT8" for "precision_mode" parameter.
     //@@
     //@@       For "tensorrt", the following parameters can be specified:
```
tanmayv25, line 906: All the supported values for ORT backend are documented here:
https://github.com/triton-inference-server/onnxruntime_backend#onnx-runtime-with-tensorrt-optimization

Below does not capture all the parameters. The supported parameters are going to be backend-specific and might change frequently. It would be better to point the users to specific backend repo README to learn about supported parameters.

rmccorm4: It doesn't look like there's any mention of TF-TRT / TRT optimization in TF backend readme: https://github.com/triton-inference-server/tensorflow_backend.

1. Maybe we can add a small section to TF backend README that simply points to: https://github.com/triton-inference-server/server/blob/main/docs/optimization.md#tensorflow-with-tensorrt-optimization-tf-trt
2. And we can add some notes to the the [TF-TRT section](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md#tensorflow-with-tensorrt-optimization-tf-trt) if we need to update any supported/unsupported fields

PR #80, opened 2023-01-19
Add protobuf for GRPC health check

dyastremsky:
Add GRPC's standard health check protocol. This can be used with tools like grpc-health-probe and Kubernetes liveness probes (as of [Kubernetes 1.24](https://kubernetes.io/blog/2022/05/13/grpc-probes-now-in-beta/)).

Related server PR (implementation, tests): https://github.com/triton-inference-server/server/pull/5267

### DIFF
--- a/protobuf/CMakeLists.txt
+++ b/protobuf/CMakeLists.txt
@@ -183,25 +183,25 @@ if(${TRITON_COMMON_ENABLE_GRPC})
   )
 
   add_library(
-    health-library EXCLUDE_FROM_ALL OBJECT
+    grpc-health-library EXCLUDE_FROM_ALL OBJECT
     ${HEALTH_SRCS} ${HEALTH_HDRS}
   )
 
   target_include_directories(
-    health-library
+    grpc-health-library
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
   )
 
   target_compile_options(
-    health-library PRIVATE
+    grpc-health-library PRIVATE
     $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>:
       -Wall -Wextra -Wno-unused-parameter -Werror>
     $<$<CXX_COMPILER_ID:MSVC>:/W0 /D_WIN32_WINNT=0x0A00 /EHsc>
   )
 
   set_target_properties(
-    health-library
+    grpc-health-library
     PROPERTIES
       POSITION_INDEPENDENT_CODE ON
   )
--- a/protobuf/CMakeLists.txt
+++ b/protobuf/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2021-2023, NVIDIA CORPORATION. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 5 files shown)


protobuf/health.proto
```diff
@@ -0,0 +1,85 @@
...
+  //@@       (HealthCheckResponse)
+  //@@
+  //@@     Get serving status of the inference server.
+  //@@
+  rpc Check(HealthCheckRequest) returns (HealthCheckResponse);
+
+  //@@  .. cpp:var:: rpc Watch(HealthCheckRequest) returns
+  //@@       (stream HealthCheckResponse)
+  //@@
+  //@@     Get current serving status of the inference server and
```
tanmayv25, line 81: What do you mean by future status changes? Can you elaborate about `Watch()` function? How does `SERVICE_UNKNOWN` used in context of `Watch()`?

I wonder if the function is not implemented, then why should it be included in the RPC proto.

dyastremsky: I'll remove it from the RPC proto. I was trying to keep the original protobuf intact, as described in the [GRPC health checking documentation](https://github.com/grpc/grpc/blob/master/doc/health-checking.md), and leave it for users to see. Plus, it'd make implementing it in the future easier if we decide to. But you're right, we can remove it from the proto.

In case you're still curious about Watch, the official protobuf actually has more detailed comments [here](https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto#L47-L62) that we could comment if we choose to implement it in the future.

dyastremsky: Done. Let me know if that works for you, @tanmayv25.

tanmayv25 (approved):

protobuf/CMakeLists.txt
```diff
@@ -160,3 +160,77 @@ if(${TRITON_COMMON_ENABLE_GRPC})
...
+      OUTPUT "${HEALTH_PY}"
+      COMMAND ${Python_EXECUTABLE}
+      ARGS
+        -m grpc_tools.protoc
+        -I "${health_proto_dir}"
+        --grpc_python_out "${CMAKE_CURRENT_BINARY_DIR}"
+        "health.proto"
+      DEPENDS "health.proto" proto-library
+    )
+  endif()
```
Tabrizian, line 200: Is the Python grpc file required? I believe we only need to generate if we want to use it in the Python clients.

protobuf/health.proto
```diff
@@ -0,0 +1,77 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto3";
+
+package grpc.health.v1;
```
Tabrizian, line 29: I was wondering why do you create a separate GRPC service? Is it possible to add the new RPC definitions in our existing grpc.proto?

dyastremsky: It's possible, but it'd require more work for the user. Kubernetes [allows users to specify a non-default service](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) with an extra argument. Other tools may not allow users to do so. In any case, this is how most projects implement the health service for GRPC standard and it's where/how users would expect to interact with it.

protobuf/CMakeLists.txt
dyastremsky: Good catch, that's probably not needed. Removed.

Tabrizian (approved):

PR #81, opened 2023-02-14
Add reserved namespace field in ensemble step

GuanLuo:


kthui (approved):

tanmayv25 (approved):

dyastremsky (approved):

rmccorm4 (approved):

Tabrizian (approved):

rmccorm4:
Out of curiosity, does this have to be a model config proto field? Is there any way it can be tracked in ModelInfo or any other internal structs we use after parsing model configs?

GuanLuo:
It doesn't have to but it is convenient at the moment, and it is reasonable to assume the field will be exposed to user in the future. That being said, it will be useful if we have a superset of model config for internal use

PR #82, opened 2023-03-07
Change request priority from uint32 to uint64

QMassoz:
Request priority is defined in [uint64 on client-side](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/library/common.h#:~:text=for%20the%20model.-,uint64_t,-priority_%3B) but [uint32 on server-side](https://github.com/triton-inference-server/core/blob/main/src/infer_request.h#:~:text=uint32_t%20batch_size_%3B-,uint32_t,-priority_%3B). This pull request removes this discrepancy and allows uint64 request priority on server-side. An application of this feature is to enable the use of unix timestamp as request priority.

This is linked to PR https://github.com/triton-inference-server/core/pull/176

### DIFF
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -1134,7 +1134,7 @@ message ModelDynamicBatching
   //@@
   bool preserve_ordering = 3;
 
-  //@@  .. cpp:var:: uint32 priority_levels
+  //@@  .. cpp:var:: uint64 priority_levels
   //@@
   //@@     The number of priority levels to be enabled for the model,
   //@@     the priority level starts from 1 and 1 is the highest priority.
@@ -1143,14 +1143,14 @@ message ModelDynamicBatching
   //@@     priority 3, etc. Requests with the same priority level will be
   //@@     handled in the order that they are received.
   //@@
-  uint32 priority_levels = 4;
+  uint64 priority_levels = 4;
 
-  //@@  .. cpp:var:: uint32 default_priority_level
+  //@@  .. cpp:var:: uint64 default_priority_level
   //@@
   //@@     The priority level used for requests that don't specify their
   //@@     priority. The value must be in the range [ 1, 'priority_levels' ].
   //@@
-  uint32 default_priority_level = 5;
+  uint64 default_priority_level = 5;
 
   //@@  .. cpp:var:: ModelQueuePolicy default_queue_policy
   //@@
@@ -1161,13 +1161,13 @@ message ModelDynamicBatching
   //@@
   ModelQueuePolicy default_queue_policy = 6;
 
-  //@@  .. cpp:var:: map<uint32, ModelQueuePolicy> priority_queue_policy
+  //@@  .. cpp:var:: map<uint64, ModelQueuePolicy> priority_queue_policy
   //@@
   //@@     Specify the queue policy for the priority level. The default queue
   //@@     policy will be used if a priority level doesn't specify a queue
   //@@     policy.
   //@@
-  map<uint32, ModelQueuePolicy> priority_queue_policy = 7;
+  map<uint64, ModelQueuePolicy> priority_queue_policy = 7;
 }
 
 //@@


GuanLuo:
Hi @QMassoz, sorry for the delay, really appreciate for identifying the misalignment and contributing a fix.

After reviewing the change, I think it is better to modify the client side code to align the data type on priority (uint32_t), change in the server side / GRPC service may not be backward-compatible. Do you mind to apply the change with respect to the client instead?

QMassoz:
Hi @GuanLuo, I would prefer to have uint64_t request priority rather than uint32_t because we already use this feature in production and is important for our application. 

Could you elaborate on the "change in the server side / GRPC service may not be backward-compatible"? I do not have knowledge of all triton components and how they play together, so I am interested in understanding. 
- For the GRPC service: some priority-related fields of ModelDynamicBatching have been changed in [common/protobuf/model_config.proto](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto) from uint32 to uint64. Since config.pbtxt is parsed at server launch, the model repositories should be compatible with any version of the server. What could go wrong? The only issue I could identify is if you have triton components (server, core, tensorrt-backend, etc) compiled with a different "common" component, but does it happen?
- For the triton API compatibility: the signatures of TRITONSERVER_InferenceRequestPriority and TRITONSERVER_InferenceRequestSetPriority have been changed in [core/inclure/triton/core/tritonserver.h](https://github.com/triton-inference-server/core/blob/75dbd1738078af3e669d6f893574bb9726907f2b/include/triton/core/tritonserver.h#L1079). By definition it is indeed not backward compatible and requires a bump of major version, but these functions are already called with uint64 values in other triton components (see [TRITONSERVER_InferenceRequestPriority](https://github.com/search?q=org%3Atriton-inference-server%20TRITONSERVER_InferenceRequestPriority&type=code) and [TRITONSERVER_InferenceRequestSetPriority](https://github.com/search?q=org%3Atriton-inference-server+TRITONSERVER_InferenceRequestSetPriority+&type=code)). Shouldn't it be -in practice- backward compatible since all components using these two functions are compiled alongside the core component in server/build.py ? An exception could be the C-API but I have zero knowledge on that.

GuanLuo:
> The only issue I could identify is if you have triton components (server, core, tensorrt-backend, etc) compiled with a different "common" component, but does it happen?

That is usually the concern for changing the protobuf definition / in-process API (C API) declaration. It is true that the Triton release shipped has all components compiled against the same source, however, many users do bring in their own customization (i.e. custom backend / frontend against `libtritonserver.so`), which pins to a fixed Triton interface version to simplify development. It is probably not intended to make interface changes that is not backward-compatible, can you share the commit of the change you mentioned so I can take a closer look?

I do notice that the server frontend is receiving the priority parameter as 64 bit variable which can result in overflow when calling the in-process API. I think we would like to correct that (@tanmayv25 ).

I think to fix the alignment issue, the direction should be based on the data type defined by core and span out. But the 64 bit support can well be a request for enhancement if there is a valid concern. Do you mind to share the reasoning of needing to expand priority level to be 64 bit? In my opinion, 32 bit already allows many priority levels to be established.

CC @nnshah1 @tanmayv25

QMassoz:
>  can you share the commit of the change you mentioned so I can take a closer look?

- for the protobuf definition: [a866518e8339eff57635dea42b9f2e521521e15d](https://github.com/triton-inference-server/common/pull/82/commits/a866518e8339eff57635dea42b9f2e521521e15d)
- for the triton API: [6c9b375a04ba3fe07f2989028ebe515e4e28f33d](https://github.com/triton-inference-server/core/pull/176/commits/6c9b375a04ba3fe07f2989028ebe515e4e28f33d#diff-5ed8dbbcfda4302eb8115993e5297cdf1e184f856c415059946525a7b618397c) for tritonserver.c and [55fc3cf5ee9e2aa54534d7087b452affbf5589ab](https://github.com/triton-inference-server/core/pull/176/commits/55fc3cf5ee9e2aa54534d7087b452affbf5589ab#diff-ce70b44ec4760c05dc91fc32ae46c5c3d9fa27edb61b960bf299b5960b38fc9f) for tritonserver.h

> I think to fix the alignment issue, the direction should be based on the data type defined by core and span out. But the 64 bit support can well be a request for enhancement if there is a valid concern. Do you mind to share the reasoning of needing to expand priority level to be 64 bit? In my opinion, 32 bit already allows many priority levels to be established.

I work in the field of live broadcast of sport events where there are many cameras and many replay operators. Our inference servers enable operators to apply some special effects on their clip of choice using a stream of requests (low turnaround and real-time are crucial here). 
Each operator uses a different machine and communication between machines is difficult to put into place, so we have many independent triton clients that may send requests simultaneously. Therefore, we use the UNIX timestamp as the priority level. 32 bits do allow 4M values, but it would only allow a granularity in seconds using a non-conventional epoch (see helper types [here](https://en.cppreference.com/w/cpp/chrono/duration)). 64 bits allow milliseconds-granular timestamps using the traditional epoch.

GuanLuo:
I see, so it uses priority queues to make a time prioritized scheduler. In this case it makes sense to have priority API align with common timestamp type, I would be fine to have new in-process API for setting priority of 64 bit value (i.e. `TRITONSERVER_InferenceRequestSetPriorityUint64`?). It is a bit tricky because then there are two sets of getter/setter for different range and overflow can happen if mix of getter / setter is used. @nnshah1  please advise on API changes

GuanLuo:
@QMassoz While there will be some more brainstorming on the API change, I wonder if the following can be a possible WAR for the moment: you will still retrieve timestamp at millisecond granularity (uint64_t) and down-cast / bit-mask it to uint32_t as the request priority? Then you can still preserve the millisecond-level detail, with the cost that the range can only cover ~ 50 days.

nnshah1:
> I see, so it uses priority queues to make a time prioritized scheduler. In this case it makes sense to have priority API align with common timestamp type, I would be fine to have new in-process API for setting priority of 64 bit value (i.e. `TRITONSERVER_InferenceRequestSetPriorityUint64`?). It is a bit tricky because then there are two sets of getter/setter for different range and overflow can happen if mix of getter / setter is used. @nnshah1 please advise on API changes

@QMassoz, @GuanLuo 

Thanks for bringing this up - it looks like currently the protocol specification and implementation are little different on this point (grpc inference request uses int64, in process api uses uint32, http json uses uint) - I'll work with @GuanLuo and team on a proposal. 

@QMassoz From what we understand you aren't blocked at the moment - is that right?

nnshah1:
@QMassoz We have added the support for uint64 request priorities both in the model config and in the request parameters. Please let us know if anything is missing for your use case.

PR #84, opened 2023-03-28
Add new JSON dt checks with testing

Tabrizian:
Server PR change: https://github.com/triton-inference-server/server/pull/5560


include/triton/common/triton_json.h
```diff
@@ -915,6 +915,42 @@ class TritonJson {
       return TRITONJSON_STATUSSUCCESS;
     }
 
+    bool IsString()
+    {
+      const rapidjson::Value& object = AsValue();
+      return object.IsString();
```
GuanLuo, line 921: Is there way to get the internal type information of the object?

GuanLuo (approved):

kthui (approved):

krishung5 (approved):

tanmayv25 (approved):

include/triton/common/triton_json.h
Tabrizian: There is an internal type. The problem is that rapidjson uses different types for true and false: https://rapidjson.org/namespacerapidjson.html#ae79a4751c1c460ff0de5ecc07874f3e4 It means that we would have to create another wrapper that maps these to the boolean type.

GuanLuo: I think if we were going to expose the value type, then we would have the "Triton JSON type" wrapper around the rapidjson type anyway, the same as the Triton JSON class encapsulates the rapidjson details.

PR #85, opened 2023-04-03
Add rapidjson include headers for test

Tabrizian:


src/test/CMakeLists.txt
```diff
@@ -37,4 +37,7 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(googletest)
 
-add_subdirectory(triton_json triton_json)
+if (TRITON_COMMON_ENABLE_JSON)
+    add_subdirectory(triton_json triton_json)
```
mc-nv, line 41: I suppose we need to change option here to be able use this condition.
https://github.com/triton-inference-server/common/blob/main/CMakeLists.txt#L40

GuanLuo: I wonder if that has been turned off explicitly in FIL backend build.

Tabrizian: This was on in the FIL backend. Not sure why it was only failing in case of FIL backend.

Tabrizian: @mc-nv This is a parameter that the user can set ON/OFF if it needs TritonJson or not.

PR #86, opened 2023-04-10
Fix rapidjson build in windows

tanmayv25:
re2/stringpiece.h uses std::min/max which seems to be breaking rapidjson build on windows.

https://github.com/google/re2/blob/2022-06-01/re2/stringpiece.h#L141

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,8 @@ endif()
 
 add_library(common-compile-settings INTERFACE)
 
+target_compile_definitions(common-compile-settings INTERFACE NOMINMAX)
+
 target_compile_features(common-compile-settings INTERFACE cxx_std_11)
 
 target_compile_options(common-compile-settings INTERFACE
--- a/include/triton/common/triton_json.h
+++ b/include/triton/common/triton_json.h
@@ -30,6 +30,8 @@
 // RapidJSON's GetObject.
 // https://github.com/Tencent/rapidjson/issues/1448
 #undef GetObject
+// Define NOMINMAX to build successfully with std::min/max
+#define NOMINMAX
 #include <rapidjson/document.h>
 #else
 // Disable class-memaccess warning to facilitate compilation with gcc>7
--- a/include/triton/common/triton_json.h
+++ b/include/triton/common/triton_json.h
@@ -30,8 +30,6 @@
 // RapidJSON's GetObject.
 // https://github.com/Tencent/rapidjson/issues/1448
 #undef GetObject
-// Define NOMINMAX to build successfully with std::min/max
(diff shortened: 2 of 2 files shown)


dyastremsky (dismissed):

include/triton/common/triton_json.h
```diff
@@ -30,6 +30,8 @@
 // RapidJSON's GetObject.
 // https://github.com/Tencent/rapidjson/issues/1448
 #undef GetObject
+// Define NOMINMAX to build successfully with std::min/max
+#define NOMINMAX
```
nv-kmcgill53, line 34: Are we able to define this in cmake so that we don't have to define this in every header file?  I don't believe we use the `MIN`/`MAX` macros from windows at all.

nv-kmcgill53 (changes requested):

include/triton/common/triton_json.h
tanmayv25: I have created a follow-up ticket to explore this options: DLIS-4769.

tanmayv25: Looks like there is still some std::min/max being included in the library. I am going to experiment with the compiler flags next.

dyastremsky: Windows.h is probably included earlier (especially given that undef GetObject is called above, which would be called after windows.h per the comments)? You'd need to include it before windows.h is included.

nv-kmcgill53: @dyastremsky Correct, we don't have the issue until `windows.h` is included. If we are able to define `NOMINMAX` on the compilation line then that define is defined at the start of every header.

CMakeLists.txt
```diff
@@ -55,6 +55,8 @@ endif()
 
 add_library(common-compile-settings INTERFACE)
 
+target_compile_definitions(common-compile-settings INTERFACE NOMINMAX)
```
nv-kmcgill53, line 58: nit: Are we able to put this inside an if-else for windows?

```suggestion
if (WIN32)
     target_compile_definitions(common-compile-settings INTERFACE NOMINMAX)
endif()
```

nv-kmcgill53 (changes requested):

include/triton/common/triton_json.h
tanmayv25: Thanks for the catch!

CMakeLists.txt
nv-kmcgill53: My mistake. Thought you were trying to get this in tonight.

tanmayv25:
Should be fixed with: https://github.com/triton-inference-server/server/pull/5625

PR #91, opened 2023-05-31
Updating Service and Model Config Protobuf for uint64 Request Priority

nnshah1:
Updating priority levels within model config proto to support uint64 (allows for utilizing timestamps as priority). 
See discussion here https://github.com/triton-inference-server/common/pull/82#issuecomment-1547650136.

In addition update the InferParameter choices to include double and uint64.

Double is to keep consistent with other inference servers - in anticipation of KServe protobuf's changes.

### DIFF
--- a/protobuf/grpc_service.proto
+++ b/protobuf/grpc_service.proto
@@ -462,6 +462,18 @@ message InferParameter
     //@@       A string parameter value.
     //@@
     string string_param = 3;
+
+    //@@    .. cpp:var:: double double_param
+    //@@
+    //@@       A double parameter value.
+    //@@
+    double double_param = 4;
+
+    //@@    .. cpp:var:: uint64 uint64_param
+    //@@
+    //@@       A uint64 parameter value.
+    //@@
+    uint64 uint64_param = 5;
   }
 }
 
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -1134,7 +1134,7 @@ message ModelDynamicBatching
   //@@
   bool preserve_ordering = 3;
 
-  //@@  .. cpp:var:: uint32 priority_levels
+  //@@  .. cpp:var:: uint64 priority_levels
   //@@
   //@@     The number of priority levels to be enabled for the model,
   //@@     the priority level starts from 1 and 1 is the highest priority.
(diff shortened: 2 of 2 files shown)


include/triton/common/triton_json.h
```diff
@@ -239,6 +239,15 @@ class TritonJson {
...
 
+    // Set/overwrite an unsigned integer in a value. This changes the
+    // type of the value to unsigned int.
+    TRITONJSON_STATUSTYPE SetUInt(const uint64_t value)
+    {
+      rapidjson::Value& v = AsMutableValue();
+      v.SetUint64(value);
+      return TRITONJSON_STATUSSUCCESS;
+    }
+
```
rmccorm4, line 250: 1. Is a SetDouble needed to go with `double_param` addition?
2. For the `// FIXME` comment a few lines up, maybe it can be removed?

protobuf/grpc_service.proto
```diff
@@ -462,6 +462,18 @@ message InferParameter
     //@@       A string parameter value.
     //@@
     string string_param = 3;
+
+    //@@    .. cpp:var:: double double_param
+    //@@
+    //@@       A double parameter value.
+    //@@
+    double double_param = 4;
```
Tabrizian, line 470: I think it might be better to add this parameter when `double` is added to the kserve protocol. Looks like currently it is not being used and we don't have any testing for it so I think it would be better to remove it for now.

rmccorm4: I was thinking the same, but wasn't sure if it was explicitly agreed to be added from kserve meeting or not.

include/triton/common/triton_json.h
nnshah1: I've added the SetDouble - though I don't believe we use it anywhere at the moment. I've removed the comment - although we don't mirror all "set" types - I don't think we actually need to so think its safe to expand as we need to.

protobuf/grpc_service.proto
nnshah1: it is agreed to be added - I'm fine to remove it until agreed upon - only thing is we will keep the uint64 at # 5 to avoid conflict. Overall preference?

rmccorm4: > only thing is we will keep the uint64 at # 5 to avoid conflict.
conflict with who? Are double/uint64 already defined as 4/5 in kserve?

I don't see it in kserve/open-inference-protocol repos. Can you share link to conflict?

nnshah1: https://github.com/Xilinx/inference-server/blob/663098bc7320d345fc3676aa8b7d361814352c37/src/amdinfer/core/inference.proto#L17

I plan to make a consolidated change to KServe after our changes are merged

tanmayv25 (approved):

Tabrizian (approved):
Approving, I think the PR for adding support for double/uint64 generic parameter support should be included in 23.07.

PR #93, opened 2023-06-08
Updating Service and Model Config Protobuf for uint64 Request Priority

nnshah1:
Updating priority levels within model config proto to support uint64 (allows for utilizing timestamps as priority).
See discussion here https://github.com/triton-inference-server/common/pull/82#issuecomment-1547650136.

In addition update the InferParameter choices to include double and uint64.

Double is to keep consistent with other inference servers - in anticipation of KServe protobuf's changes.

### DIFF
--- a/include/triton/common/triton_json.h
+++ b/include/triton/common/triton_json.h
@@ -239,6 +239,15 @@ class TritonJson {
       return TRITONJSON_STATUSSUCCESS;
     }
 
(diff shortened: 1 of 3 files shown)


rmccorm4 (approved):

rmccorm4 (approved):
This is a re-merge of https://github.com/triton-inference-server/common/pull/91 after briefly reverting it. Referencing for historical lookback purposes to see discussions.

PR #94, opened 2023-06-14
Update clang format version

dyastremsky:
The build container is now built on Ubuntu 22.04, which has clang-format-11 through clang-format-15. As a result, I'd like to update the clang format version so that we can continue to use it in the build container.

This is tested locally. When run on already-formatted files, it formats minimally. It does look like 15 has a few changes from clang-format-6.0 but the format generally makes sense.

### DIFF
--- a/tools/format.py
+++ b/tools/format.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 
-# Copyright (c) 2018-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2018-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


tools/format.py
```diff
@@ -65,7 +65,7 @@ def visit(path):
                                          style_config='google')
         return True
     else:
-        args = ['clang-format-6.0', '--style=file', '-i']
+        args = ['clang-format-15', '--style=file', '-i']
```
rmccorm4, line 68: I don't have anything against upgrading, but should we run a format sweep through all the repos and run it through CI for sanity if we're going to?

Note for others: If your dev desktop it still on Ubuntu <=20.04, you'll need to add LLVM repos to apt sources to get clang 15: https://stackoverflow.com/a/66654073

nnshah1 (approved):
Approved with the caveat that we should find a way to run in CI for sanity

tools/format.py
dyastremsky: Sure, that's a great idea. Re: 20.04, we can upgrade to clang-format-12 (latest version supported by both versions of Ubuntu) if we'd prefer to go that route for compatibility reasons. Figured this is mostly used internally so having the latest version is best, but that'd leave support for 20.04 users until the next Linux upgrade.

dyastremsky:
Passed CI. Will open formatting PRs shortly.

PR #96, opened 2023-06-23
Add GitHub action to format and lint code

dyastremsky:
This pull request adds a pre-commit to run the following:

- Black for formatting Python
- Flake8 for linting Python for conformance to PEP8, PyFlakes, and circular complexity (max-line-length set to match Black's 88 default/recommended character limit; this and other settings from Black's documentation [here](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) to avoid enforcement of limit when unnecessary).
- isort for deterministically sorting and organizing Python imports
- Clang for formatting C++, C, Java, TextProto, Proto, and CUDA
- Codespell for finding common spelling mistakes

In addition, this adds these native pre-commit hooks, details [here](https://pre-commit.com/hooks.html):
- check-case-conflict
- check-executables-have-shebangs
- check-merge-conflict
- check-json
- check-toml
- check-yaml
- check-shebang-scripts-are-executable
- end-of-file-fixer
- mixed-line-ending
- requirements-txt-fixer
- trailing-whitespace

To run these locally, you can install pre-commit via `pip install pre-commit`, then go into the repo and run `pre-commit install`. After that has been done once, you just need to run `pre-commit run --all-files` any time you want to apply them. Once installed, it should run automatically on commit for files included in a commit. To correct spelling errors found by codespell when possible, you can install codespell (`pip install codespell` and potentially `pip install --upgrade codespell` to use the toml file in the directory) and call `codespell -w <path>`.

This pull request also adds a GitHub action so that these are run for every pull request and push to main. The core changes are in pre-commit.yml, pre-commit-config.yaml, and pyproject.toml. This also removes the old formatter (.clang-format, tools/format.py, tools/pre-commit). The rest of the changes are the effect of applying the action to the current repo.

Note: When this is merged, we'll want to modify the [CONTRIBUTING](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md) document in the server repository at the same time to update the instructions on running the formatter.


pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+# this is only to allow you to run codespell interactively
+skip = "./.git,./.github"
+# ignore short words, and typename parameters like OffsetT
+ignore-regex = "\\b(.{1,4}|[A-Z]\\w*T)\\b"
+# use the 'clear' dictionary for unambiguous spelling mistakes
+builtin = "clear"
+# disable warnings about binary files and wrong encoding
+quiet-level = 3
+
+[tool.isort]
```
csadorf, line 38: The `isort` tool comes with a built-in ["black" profile](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html). You might have reason to deviate from that, but just pointing out that it exists.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,73 @@
...
+- repo: https://github.com/psf/black
+  rev: 23.1.0
+  hooks:
+      - id: black
+        types_or: [python, cython]
+- repo: https://github.com/PyCQA/flake8
+  rev: 5.0.4
+  hooks:
+      - id: flake8
+        args: [--max-line-length=88, --select=C,E,F,W,B,B950, --extend-ignore = E203,E501]
```
csadorf, line 42: Maybe it would be better to specify those flags in the `pyproject.toml` file? In this way running flake8 manually or through an IDE will yield consistent results.

csadorf (commented):
Very nice! Just a few minor comments.

You mention in the PR description that it would be advisable to adjust the CONTRIBUTING.md. I agree, but why not modify the file directly in this PR?

pyproject.toml
dyastremsky: Thanks! I didn't know, will add it.

.pre-commit-config.yaml
dyastremsky: Great point, will update this.

dyastremsky:
@csadorf Thanks for your comments! The rationale for not updating the CONTRIBUTING document here is because it's in the triton-inference-server/server repo. When we roll these changes out to all repos next week, we'll update the CONTRIBUTING document in the server repo in the same PR.

.pre-commit-config.yaml
dyastremsky: I'm actually going to revert this. Flake8 [doesn't support pyproject.toml](https://github.com/PyCQA/flake8/issues/234), and we don't currently use Flake8 standalone. So to keep from using an unofficial wrapper and adding a dependency, I'll move the command line back to the pre-commit-config.yaml file.

.clang-format
```diff
@@ -34,4 +35,4 @@ BinPackArguments: true
 BinPackParameters: true
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
 
-IndentCaseLabels: true
```
fpetrini15, line 37: Nit: newline

dyastremsky: There should be a newline. :) I think Git only shows the lack of a newline (see red/removed area vs. green/added area).

fpetrini15 (approved):
I was going to ask whether adding "black" as a profile in the pyproject.toml would impact the other PRs, but it looks like you've already updated them. Great work!

PR #104, opened 2023-11-06
Add options for growable memory and single state buffers

Tabrizian:


protobuf/model_config.proto
```diff
@@ -1382,6 +1382,24 @@ message ModelSequenceBatching
...
+    //@@  .. cpp:var:: bool use_single_buffer
+    //@@
+    //@@     The optional field to use a single buffer for both input and output state.
+    //@@     The default value is false.
+    //@@
+    bool use_single_buffer = 6;
+
+    //@@  .. cpp:var:: bool use_growable_memory
+    //@@
+    //@@     The optional field to allow an implicit state buffer to grow or shrink
```
rmccorm4, line 1395: "or shrink" - from our chat I don't think it can shrink. Are there plans to shrink in the near future? Or should we just remove it for now?

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,24 @@ message ModelSequenceBatching
...
+    //@@     when the size changes during a sequence. When using this option Triton
+    //@@     guarantess that it will use the same buffer even if the state size changes.
```
rmccorm4, line 1397: "guarantees"

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,24 @@ message ModelSequenceBatching
...
+    //@@     Currently, this option only applies for implicit state that uses CUDA and
+    //@@     use_single_buffer must be enabled.
```
rmccorm4, line 1399: IIUC, we force the implicit state memory used in growable block manager to be GPU memory for now. Maybe the wording here could be clearer that regardless of what the user does, this will use GPU memory for now, even if they request other memory types.

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,24 @@ message ModelSequenceBatching
     //@@     The optional field to specify the initial state for the model.
     //@@
     repeated InitialState initial_state = 5;
+
+    //@@  .. cpp:var:: bool use_single_buffer
+    //@@
+    //@@     The optional field to use a single buffer for both input and output state.
+    //@@     The default value is false.
+    //@@
+    bool use_single_buffer = 6;
```
GuanLuo, line 1391: Can you add a few sentences on the performance impact of `use_single_buffer = true` if any? And what are the cases that user may want to pick one over the other?

protobuf/model_config.proto
GuanLuo: Does this imply the same as the data content within the changed size is preserved?

nnshah1: How about: 

"grow or shrink when the size changes with minimal copies. Triton will reuse existing allocations instead of requesting new memory if possible"

I think "same buffer" is misleading a little here - as its not the "same buffer" if the size has changed?

maybe over thinking the terminology :)

protobuf/model_config.proto
nnshah1: Why wouldn't we error if they try different types? Just double checking - we don't silently switch preferences, correct?

protobuf/model_config.proto
nnshah1, line 1391: maybe this should be more verbose:

bool use_same_buffer_for_input_output

protobuf/model_config.proto
Tabrizian: > Does this imply the same as the data content within the changed size is preserved?

Yes

Tabrizian: @nnshah1 I updated the comment. Let me know if it looks better.

protobuf/model_config.proto
Tabrizian: >Why wouldn't we error if they try different types?

What the user specifies is just "preferences" (i.e., there is no guarantee that Triton will satisfy the request). The same pattern exists in `TRITONBackend_OutputBuffer`.

protobuf/model_config.proto
Tabrizian: added

protobuf/model_config.proto
Tabrizian: removed shrink for now.

protobuf/model_config.proto
Tabrizian: updated.

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,36 @@ message ModelSequenceBatching
     //@@     The optional field to specify the initial state for the model.
     //@@
     repeated InitialState initial_state = 5;
+
+    //@@  .. cpp:var:: bool use_same_buffer_for_input_output
+    //@@
+    //@@     The optional field to use a single buffer for both input and output
+    //@@     state. Without this option, Triton uses separate buffers for input
```
nnshah1, line 1389:
```suggestion
    //@@     state. Without this option, Triton allocates separate buffers for input
```

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,36 @@ message ModelSequenceBatching
...
+    //@@     and output state which can be problematic if the state size is
+    //@@     large. This option can help reduce the memory usage when
```
nnshah1, line 1391:
```suggestion
    //@@     large. This option reduces memory usage by allocating a single buffer.
```

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,36 @@ message ModelSequenceBatching
...
+    //@@     the state size is large. There is no harm in always enabling
```
nnshah1, line 1392:
```suggestion
    //@@     There is no harm in always enabling
```

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,36 @@ message ModelSequenceBatching
...
+    //@@     this option if the output state will be written after
+    //@@     the input state is read by the framework. Note that when using this
```
nnshah1, line 1394:
```suggestion
    //@@     When enabled the state will always be updated independent of whether TRITONBACKEND_StateUpdate is called (however TRITONBACKEND_StateUpdate should still be called for completeness).
```

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,36 @@ message ModelSequenceBatching
...
+    //@@     option, `TRITONBACKEND_StateUpdate` has no effect and the state
+    //@@     will be always updated.
+    //@@     The default value is false.
+    //@@
+    bool use_same_buffer_for_input_output = 6;
+
+    //@@  .. cpp:var:: bool use_growable_memory
+    //@@
+    //@@     The optional field to allow an implicit state buffer to grow the
```
nnshah1, line 1403: Does this need to be optional? Wondering if we could always grow memory if using CUDA memory in the implementation.

nnshah1, line 1403:
```suggestion
    //@@     The optional field to enable an implicit state buffer to grow 
```

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,36 @@ message ModelSequenceBatching
...
+    //@@     buffer when the size changes during a sequence. When using this
```
nnshah1, line 1404:
```suggestion
```

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,36 @@ message ModelSequenceBatching
...
+    //@@     option Triton guarantess that it will use the same allocations
+    //@@     even if the state size changes.
+    //@@     The added size will be appended to the end of the buffer and the
+    //@@     will be preserved. Currently, this option only applies for
+    //@@     implicit state that uses CUDA memory and use_single_buffer must be
+    //@@     existing data enabled.
+    //@@
+    //@@     The default value is false.
+    //@@
+    bool use_growable_memory = 7;
```
nnshah1, line 1414: How about the following complete description:

```
The optional field to enable an implicit state buffer to grow without reallocating and copying existing memory. Additional memory will be appended to the end of the buffer and existing data will be preserved. This option is available for CUDA memory only.
```

Tabrizian: I answered the questions in the core PR. Let me know if it still requires further clarification.

Tabrizian: The description looks good to me! Updated.

rmccorm4 (dismissed):

nnshah1 (approved):

protobuf/model_config.proto
nnshah1: I guess the question is which options take precedence - does the model config or the API call to get the buffer - will discuss offline

protobuf/model_config.proto
```diff
@@ -1382,6 +1382,40 @@ message ModelSequenceBatching
...
+    //@@  .. cpp:var:: bool use_growable_memory
+    //@@
+    //@@     The optional field to enable an implicit state buffer to grow
+    //@@     without reallocating or copying existing memory.
+    //@@     Additional memory will be appended to the end of the buffer and
+    //@@     existing data will be preserved.
+    //@@     This option is only available for CUDA memory and requires enabling
+    //@@     use_same_buffer_for_input_output. When using this option,
+    //@@     StateBuffer call will always return CUDA memory even if CPU memory
+    //@@     is provided.
```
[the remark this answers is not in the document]
Tabrizian: good catch fixed!

nnshah1 (approved):

PR #107, opened 2023-11-17
Generative ->Iterative

nnshah1:


protobuf/model_config.proto
```diff
@@ -1568,13 +1568,14 @@ message ModelSequenceBatching
...
-  //@@     batcher expects the same request to be "rescheduled" by the model if
-  //@@     the sequence is continuing.
-  bool generative_sequence = 6;
+  //@@  .. cpp:var:: bool iterative_sequence
+  //@@
+  //@@     Requests for iterative sequences are processed over a number
+  //@@     of iterations. An iterative sequence is initiated by a single
+  //@@     request and is "rescheduled" by the model until completion.
+  //@@     Requests for inflight requests will be batched together
+  //@@     and can complete independently.
```
rmccorm4, line 1577: Can we add an asterisk along the linesof "if the model/backend supports it"?

protobuf/model_config.proto
```diff
@@ -1568,13 +1568,15 @@ message ModelSequenceBatching
...
-  //@@     batcher expects the same request to be "rescheduled" by the model if
-  //@@     the sequence is continuing.
-  bool generative_sequence = 6;
+  //@@  .. cpp:var:: bool iterative_sequence
+  //@@
+  //@@     Requests for iterative sequences are processed over a number
+  //@@     of iterations. An iterative sequence is initiated by a single
+  //@@     request and is "rescheduled" by the model until completion.
+  //@@     Requests for inflight requests will be batched together
+  //@@     and can complete independently. Note this is a beta feature
```
Tabrizian, line 1577: As discussed I don't think we can have a beta model configuration parameter.

protobuf/model_config.proto
```diff
@@ -1568,13 +1568,15 @@ message ModelSequenceBatching
...
+  //@@     and requires backend support.
```
Tabrizian, line 1578: Would be good to mention the default value (i.e., false)

Tabrizian (approved):

rmccorm4 (approved):

PR #115, opened 2024-04-19
json log format

nnshah1:


src/logging.cc
```diff
@@ -66,82 +74,133 @@ Logger::Flush()
...
-  gLogger_.Log(stream_.str());
-}
+  //  gLogger_.SetLogFormat(Logger::Format::kJSONL);
 
+  switch (gLogger_.LogFormat()) {
+    case Logger::Format::kDEFAULT:
+    case Logger::Format::kISO8601: {
+      std::stringstream preamble;
+      LogPreamble(preamble);
+      //     std::string escaped(message_.str());
```
rmccorm4, line 183:
```suggestion
```

include/triton/common/triton_json.h
```diff
@@ -109,6 +110,19 @@ class TritonJson {
...
+  static std::string EscapeString(const std::string& input)
+  {
+    WriteBuffer writebuffer;
+    const unsigned int writeFlags = rapidjson::kWriteNanAndInfFlag;
+    // Provide default template arguments to pass writeFlags
+    rapidjson::Writer<
+        WriteBuffer, rapidjson::UTF8<>, rapidjson::UTF8<>,
+        rapidjson::CrtAllocator, writeFlags>
+        writer(writebuffer);
+    writer.String(input.c_str());
```
rmccorm4, line 122: I was going to ask if this can fail if the input has some malformed or half-formed JSON in it, but it seems like it won't: https://github.com/Tencent/rapidjson/issues/1767#issuecomment-672541097

src/logging.cc
nnshah1: left for easy debug until we settle on direction - but good to keep track

include/triton/common/triton_json.h
nnshah1: error handling was a little tricky w.r.t to we usually log an error - so I didn't want to create a recursion by accident. Writing and error causes an error .... so would probably have to fail silently ... or print directly to the stdout ...

include/triton/common/logging.h
```diff
@@ -125,7 +129,10 @@ class Logger {
   // Flush the log.
   void Flush();
 
+  static const std::array<const char*, Level::kINFO + 1> LEVEL_NAMES;
```
GuanLuo, line 132: Should have defined `Level:kEND` then use it instead of `Level::kINFO + 1`, easy to miss if someone adds new log level and doesn't change there.

nnshah1:
continued in new PR

PR #119, opened 2024-05-23
Add support for building on Ubuntu 24.04

jmblixt3:
Add include for cstdint in error.h

### DIFF
--- a/include/triton/common/error.h
+++ b/include/triton/common/error.h
@@ -26,6 +26,7 @@
 #pragma once
 
 #include <string>
+#include <cstdint>
 
 namespace triton { namespace common {
 


include/triton/common/error.h
```diff
@@ -26,6 +26,7 @@
 #pragma once
 
 #include <string>
+#include <cstdint>
```
francis-gagnon, line 29: Tested on Fedora 40 with gcc version 14.1.1 and it is indeed necessary to add this inclusion for compilation.

francis-gagnon (approved):

iamthebot (approved):
Tested this patch works on Ubuntu 22.04 with GCC 14.1.1.

This is probably going to be required for all newer distros as well.

rmccorm4:
Hi @jmblixt3 , thanks for the contribution!
 
@mc-nv @nv-kmcgill53 @fpetrini15 can you take a sanity check look here to make sure this is OK with RHEL, Windows, etc?

fpetrini15:
@rmccorm4 I just tested compiling a simple program with this include in our RHEL base and Windows build base container and didn't encounter any issues.

rmccorm4:
Seems to be something with gcc 14 specifically, it looks like our builds use gcc 11. Similar PR: https://github.com/NVIDIA/DALI/issues/5591

rmccorm4 (approved):

nv-kmcgill53:
I don't understand why this is a breaking change with a newer version of gcc. That seems like bad design. Approved.

nv-kmcgill53 (approved):

nv-kmcgill53:
@jmblixt3 Have you filled out the Triton CCLA? I don't see your company or or user in our files. If not, can you please fill this out so we can merge your PR? Thanks!

https://github.com/triton-inference-server/server/blob/main/Triton-CCLA-v1.pdf

jmblixt3:
> @jmblixt3 Have you filled out the Triton CCLA? I don't see your company or or user in our files. If not, can you please fill this out so we can merge your PR? Thanks!
> 
> https://github.com/triton-inference-server/server/blob/main/Triton-CCLA-v1.pdf

Just did it

nv-kmcgill53:
Thank you for sending in your CLA and contributing to the Triton project. We have accepted it and can merge your PR at our discretion.

PR #120, opened 2024-06-12
fix: guard windows api definitions in logging.h

nnshah1:
The common logging library is compiled and incorporated as a static library. When including the logging.h file in other projects the winsock apis can conflict and thus we need to guard against including them. 

This is a similar pattern to: 

https://github.com/triton-inference-server/server/blob/ab5bedc63752f4fdfab4c0d8bd8ddf4630e76573/src/main.cc#L29

### DIFF
--- a/include/triton/common/logging.h
+++ b/include/triton/common/logging.h
@@ -36,9 +36,9 @@
 
 #include "table_printer.h"
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


rmccorm4 (approved):
LGTM and I was able to get a successful build locally with these changes.

Waiting for pipeline 15817551 job 97434731 to confirm this works in CI as well.

PR #125, opened 2024-07-14
Fix shape and reformat free tensor handling in the input byte size check

pskiran1:
- Added support for the `is_non_linear_format_io` flag in the model configuration to identify if input/output tensors use a non-linear IO format.

### DIFF
--- a/protobuf/model_config.proto
+++ b/protobuf/model_config.proto
@@ -410,14 +410,14 @@ message ModelInput
   //@@
   bool optional = 8;
 
-  //@@  .. cpp:var:: bool is_reformat_free_tensor
+  //@@  .. cpp:var:: bool is_non_linear_format_io
   //@@     
-  //@@     Whether or not the input is a reformat-free tensor to the model. This
-  //@@     field is currently supported only for the TensorRT model. An error 
-  //@@     will be generated if this specification does not comply with underlying
+  //@@     Indicates whether the input tensor uses a non-linear IO format. This
+  //@@     field is currently supported only for TensorRT models. An error will 
+  //@@     be generated if this specification does not comply with the underlying
   //@@     model.
   //@@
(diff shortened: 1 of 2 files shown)


protobuf/model_config.proto
```diff
@@ -409,6 +409,16 @@ message ModelInput
...
+
+  //@@  .. cpp:var:: bool allow_ragged_batch
+  //@@
+  //@@     Whether or not the input is allowed to be "ragged" in a dynamically
+  //@@     created batch. Default is false indicating that two requests will
+  //@@     only be batched if this tensor has the same shape in both requests.
+  //@@     True indicates that two requests can be batched even if this tensor
+  //@@     has a different shape in each request.
+  //@@
+  bool allow_ragged_batch = 9;
```
tanmayv25, line 421: > Don’t change the field numbers for any existing fields. “Changing” the field number is equivalent to deleting the field and adding a new field with the same type. If you want to renumber a field, see the instructions for [deleting a field](https://protobuf.dev/programming-guides/proto3/#deleting).

Source: https://protobuf.dev/programming-guides/proto3/#updating

tanmayv25: See how the kind field number is different. The order of the tag number does not matter. In this case, `allow_ragged_batch` should be 7 and the `is_non_linear_format_io` should be 9.

pskiran1: Sure, thank you. Could you please confirm if it is possible to use different numbers for the same field name in the input and output?
input: is_non_linear_format_io=7
output: is_non_linear_format_io=9

pskiran1: Made the change mentioned above and and have not found any issues in the CI.

tanmayv25 (approved):

GuanLuo (approved):

PR #131, opened 2025-02-26
Provide preview of JSON when Parse fails

dyastremsky:
Parse is used across the Triton codebase. When Parse is failed, all the user sees is the error type and index. To help with debugging, provide a preview of the parsed JSON.


include/triton/common/triton_json.h
```diff
@@ -194,6 +200,31 @@ class TritonJson {
...
+        error_stream << "failed to parse the request JSON buffer: "
+                     << GetParseError_En(document.GetParseError())
+                     << " at offset " << document.GetErrorOffset() << ".";
+
+        // Show part of the JSON to help debugging
+        const size_t preview_length = 100;
+        std::string json_preview = json.substr(0, preview_length);
+        if (json.size() > preview_length) {
+          json_preview += "...";
+        }
```
rmccorm4, line 219: Any strong reason to not just include the whole thing?

For example, can the error start at some point after the preview length?

```{... /* 100+ chars */, "my_invalid_json".123}```

File: include/triton/common/triton_json.h [hunk 2]
```diff
@@ -184,6 +184,12 @@ class TritonJson {
             std::string(GetParseError_En(document_.GetParseError())) + " at " +
             std::to_string(document_.GetErrorOffset())));
       }
+      TRITONJSON_STATUSTYPE status =
+          ParseErrorHandler(document_, std::string(base, size));
```
rmccorm4, line 188: Can you add a before/after example in the PR description to highlight the new error/UX improvement?

dyastremsky: Sure! Let me run it through CI then. I wanted your feedback first in case there was a reason not to do this.

include/triton/common/triton_json.h
dyastremsky: I wanted to avoid that in case the JSON might be very long and crashes the program. I can increase the preview length to the limit of your choice (maybe `std::numeric_limits<std::size_t>::max()`or print the whole thing. Any preference?

dyastremsky: That is a fair point that it could very well be an issue after the 100th character. A large limit could resolve that while balancing the risk of the JSON being too long and causing a crash.

PR #134, opened 2025-03-24
fix: Prevent infinite loop when terminal column size is set to a very low value

pskiran1:
This change primarily ensures that `terminal_width` and `equal_share` are positive values, preventing `shares_` from becoming excessively large in later calculations.


indrajit96 (approved):
LGTM!
Please attach steps to repro, RCA and issue getting resolved after the fix in the PR.

yinggeh (approved):

PR #157, opened 2026-04-24
feat: Add `.MemberIsNullIs[type] Functions to TritonJson

whoisj:
This change adds a set of` .MemberIs[type]` functions to TritonJson.

* `MemberIsArray(const char* name) const`
* `MemberIsArray(const std::string& name) const`
* `MemberIsNull(const char* name) const`
* `MemberIsNull(const std::string& name) const`
* `MemberIsObject(const char* name) const`
* `MemberIsObject(const std::string& name) const`
* `MemberIsString(const char* name) const`
* `MemberIsString(const std::string& name) const`

These new methods are used in parsing of the JSON blobs returned by PyTorch's AOTInductorContainerRunner::get_call_spec() function to determine the interface the model supports.

### DIFF
--- a/include/triton/common/triton_json.h
+++ b/include/triton/common/triton_json.h
@@ -1003,8 +1003,9 @@ class TritonJsonImpl {
     bool MemberIsString(const char* name) const
     {
       const rapidjson::Value& object = AsValue();
-      if (!object.IsObject() || !object.HasMember(name))
+      if (!object.IsObject() || !object.HasMember(name)) {
         return false;
+      }
       const auto& v = object[name];
       return v.IsString();
     }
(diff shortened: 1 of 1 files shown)


yinggeh (dismissed):
LGTM

include/triton/common/triton_json.h
```diff
@@ -961,8 +961,9 @@ class TritonJsonImpl {
     bool MemberIsArray(const char* name) const
     {
       const rapidjson::Value& object = AsValue();
-      if (!object.IsObject() || !object.HasMember(name))
+      if (!object.IsObject() || !object.HasMember(name)) {
```
yinggeh, line 964: Good catch. We should never do no-block one-liner.

yinggeh (approved):

PR #158, opened 2026-06-25
feat: Support a log callback in the common Logger

pskiran1:
Adds an optional callback to the shared `Logger` so an embedding application can receive each log record as structured fields (level, file, line, timestamp, message) instead of parsing Triton's stdout/stderr text. When a callback is set, records go to it and the default sink is skipped. When it is not set, behavior is unchanged.

CI: https://github.com/triton-inference-server/server/pull/8858


include/triton/common/logging.h
```diff
@@ -140,6 +143,37 @@ class Logger {
     return std::string();
   }
 
+  // Delivers each enabled log record as structured fields to an embedding host
+  // (e.g. Dynamo), bypassing Triton's default stderr/file sink entirely.
+  //
+  // The callback is invoked on the producing thread, outside the output mutex.
+  // It must be lightweight, thread-safe, and must never throw.
+  // Register before the server starts serving, registration is process-global.
+  using LogCallbackFn = std::function<void(
```
whoisj: I would prefer to see this as a `typedef` as usings behave oddly when included.

`typedef std::function<void(Level level, bool is_verbose, const char* file, int line, uint64_t timestamp_us, const char* message)> LogCallBackFn;`

include/triton/common/logging.h
```diff
@@ -156,6 +190,9 @@ class Logger {
   std::mutex mutex_;
   std::string filename_;
   std::ofstream file_stream_;
+  std::atomic<bool> has_callback_{false};
```
whoisj: can't `callback_ != nullptr` provide the same functionality?

src/logging.cc
```diff
@@ -142,6 +142,37 @@ LogMessage::LogPreamble(std::stringstream& stream)
...
+  // The callback receives the unformatted message and structured metadata, and
+  // is responsible for any formatting or escaping. Invoke it outside the logger
+  // output mutex so a slow callback does not block all logging. It must never
+  // throw.
+  if (gLogger_.HasLogCallback()) {
+    Logger::LogCallbackFn callback = gLogger_.GetLogCallback();
+
+    if (callback) {
+      uint64_t timestamp_us =
+          static_cast<uint64_t>(timestamp_.tv_sec) * 1000000ULL +
```
whoisj: what is this magic number? would prefer to see it as a `constexpr`.

whoisj (changes requested):
Looks good. I have a couple of requests.

src/logging.cc
pskiran1: This converts the timestamp to a single value in microseconds for passing to Dynamo. The struct timestamp_ contains two fields (tv_sec and tv_usec), which are combined into one value.

pskiran1: Moved it to a constexpr.

include/triton/common/logging.h
pskiran1: `callback_` is protected by `callback_mutex_`. I added `has_callback_` flag as a lock-free atomic, allowing the no-callback path to perform a single inexpensive atomic read, avoiding mutex locking.

whoisj: seems like unnecessary optimization in favor of complexity.

whoisj (dismissed):
Approved. I don't think the atomic<bool> is needed, but it's also just fine.

include/triton/common/logging.h
yinggeh: Why do we even need mutex? The callback should only be set when server starts and should not change IMO.

include/triton/common/logging.h
```diff
@@ -265,6 +313,7 @@ class LogMessage {
   if (LOG_VERBOSE_IS_ON(L))                                  \
   triton::common::LogMessage(                                \
       (char*)(FN), LN, triton::common::Logger::Level::kINFO) \
+      .SetVerbose()                                          \
```
yinggeh, line 305: Why not introduce kVERBOSE?

src/test/CMakeLists.txt
```diff
@@ -39,5 +39,6 @@ FetchContent_MakeAvailable(googletest)
 
 if (TRITON_COMMON_ENABLE_JSON)
```
yinggeh, line 40: `TRITON_COMMON_ENABLE_JSON` or `TRITON_COMMON_ENABLE_LOGGING`?

include/triton/common/logging.h
pskiran1: Yes, I overlooked that. It is a process-global setting, only configured during server initialization and not allowed to be modified afterward. I have removed the mutex and the related complex checks.

src/test/CMakeLists.txt
pskiran1: The `logging_test` compiles `logging.cc`, whose default sink calls `TritonJson::SerializeString` (RapidJSON), so it needs `TRITON_COMMON_ENABLE_JSON`.
Regarding `TRITON_ENABLE_LOGGING`, it only turns the `LOG_*` macros into no-ops and it doesn't guard `logging.cc`, so the test is unaffected when it's disabled.

include/triton/common/logging.h
pskiran1: Introducing `kVERBOSE` may need large changes. Triton emits verbose at `kINFO` by design and verbose is not a separate severity. Adding `kVERBOSE` would change standalone Triton: it needs a new `LEVEL_NAMES` entry (verbose would print V… instead of I…), changes Log() routing (verbose goes to stderr instead of stdout), and mixes the `enables_` boolean with the `vlevel_` threshold model, a core change affecting all Triton users.
`SetVerbose()` keeps `level_ == kINFO`, so the default sink output is unchanged, it only tags the record so the callback can report the level accurately.

yinggeh: Got it. Also, is distinguishing verbose and info only used for `log.info` and `log.debug` purpose?

pskiran1: Yes.

include/triton/common/logging.h
yinggeh: There's no code that prevents modification after initialization. Suggestion
- Enforce it: set the callback exactly once during `TRITONSERVER_ServerNew` init (before worker/logging threads start), and either drop the public setter or make it a no-op/error after init. Then lock-free reads are genuinely safe.

whoisj (dismissed):

include/triton/common/logging.h
pskiran1: Done. I updated the code to set the callback only once `TRITONSERVER_ServerNew` init and No public APIs are exposed to set the callback during runtime or any other places. Any modifications are restricted to Options config only.

src/test/logging/logging_test.cc
```diff
@@ -0,0 +1,172 @@
...
+class LogCallbackTest : public ::testing::Test {
+ protected:
+  void TearDown() override
+  {
+    tc::gLogger_.SetLogCallback(tc::Logger::LogCallbackFn());
+  }
+};
+
+
+TEST_F(LogCallbackTest, ReceivesCorrectRecordFields)
```
yinggeh, line 92: Add a short descriptive comment for each unit test.

pskiran1: Done

yinggeh:
Fix copyrights

include/triton/common/logging.h
```diff
@@ -140,6 +142,28 @@ class Logger {
...
+      uint64_t timestamp_us, const char* message)>
+      LogCallbackFn;
+
+  // Registers or clears (pass an empty function) the log callback. When set,
+  // all log records are routed exclusively to the callback instead of
+  // the default stdout/stderr/file sink.
+  //
+  // This is set exactly once by TRITONSERVER_ServerNew, before any worker
+  // or logging threads start, and is never modified afterward.
+  void SetLogCallback(LogCallbackFn callback)
```
yinggeh: The comment says the callback is “set exactly once by TRITONSERVER_ServerNew … and is never modified afterward,” but common still exposes a public SetLogCallback / LOG_SET_CALLBACK that can clear or replace the callback at any time (and the tests do that in TearDown). So the once-only rule is a caller convention, not something this library enforces.
Please either:
- soften the comment to match the API (e.g. typically installed once at server init; must not be changed while logging is active), or
- state explicitly that common does not enforce once-only and that TRITONSERVER_ServerNew / Options owns that invariant.

pskiran1: Thank you for noting it. I updated the comment.

include/triton/common/logging.h
```diff
@@ -152,12 +152,15 @@ class Logger {
...
-  // Registers or clears (pass an empty function) the log callback. When set,
-  // all log records are routed exclusively to the callback instead of
-  // the default stdout/stderr/file sink.
+  // Registers the log callback, or clears it when passed an empty function.
+  // While set, log records are delivered only to the callback and the default
+  // stdout/stderr/file sink is bypassed.
   //
-  // This is set exactly once by TRITONSERVER_ServerNew, before any worker
-  // or logging threads start, and is never modified afterward.
+  // The read on the logging path is lock-free and this class enforces no
```
yinggeh: All lock-free code is thread-safe, but not all thread-safe code is lock-free. Not thread-safe is more appropriate.

include/triton/common/logging.h
```diff
@@ -152,12 +152,15 @@ class Logger {
...
+  // set-once policy, so the callback must be installed before any worker or
+  // logging thread starts and must not change while logging can run
+  // concurrently. In the Triton server, TRITONSERVER_ServerNew owns this and
```
yinggeh: I am confused when reading `...and must not change while logging can run concurrently`. What's the meaning of `run concurrently` here?

include/triton/common/logging.h
pskiran1: Thank you. I updated the comment.

include/triton/common/logging.h
pskiran1: I meant to specify that multiple logging message calls can run concurrently during runtime without issues. Now I updated the comment to avoid confusion.

yinggeh (approved):

src/logging.cc
```diff
@@ -142,6 +146,34 @@ LogMessage::LogPreamble(std::stringstream& stream)
...
+  // through its own logging pipeline as a single, consistent stream.
+  // The callback receives the unformatted message and structured metadata, and
+  // is responsible for any formatting or escaping. Invoke it outside the logger
+  // output mutex so a slow callback does not block all logging. It must never
+  // throw.
+  const Logger::LogCallbackFn& callback = gLogger_.LogCallback();
+  if (callback) {
+    uint64_t timestamp_us =
+        static_cast<uint64_t>(timestamp_.tv_sec) * kMicrosecondsPerSecond +
+        static_cast<uint64_t>(timestamp_.tv_usec);
```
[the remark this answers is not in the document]
pskiran1: Windows support has been deprecated

